PR #2, opened 2021-06-28
Open-source kafka library

ghost:
- [x] meta-fields filled in `package.yaml` (do not rely on package.yaml support)
- [x] Add license and changelog
- [x] version bounds for dependencies in `package.yaml`
- [x] start kafka in tests so CI can pass
- [x] no internal functions exposed
- [x] reviewed exposed constructors (maybe avoid if it would likely result in backwards-incompatible breakage in the future)
- [x] reviewed number of modules
- [x] code compiles and tests run, CI passes
- [x] `cabal haddock` reports 100% doc coverage, doc comments use haddock syntax
- [x] use open-sourced version of code in private repo succesfully


nri-kafka/src/Kafka/Test.hs
```diff
@@ -0,0 +1,251 @@
+-- | Tests, exposed so we can run them against GHCID
+module Kafka.Test
+  ( stub,
```
ghost, line 3: 💭 I think we should only expose `stub` from this public module. All the other helpers we wrote to support writing tests for this library itself. They're pretty slow and flaky and mess up test runner output. I think we might want to put them in a private module.

ghost: Going to do it.

nri-kafka/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Unreleased next version
+
+- In order to prevent accidental API changes, we force data in redis API to
+  implement the HasExamples TypeClass from the nri-test-encoding package. (thanks @stoeffel / #82)
```
ghost: Bad copy/paste from redis changelog? Will remove.

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -0,0 +1,505 @@
...
+    PartitionKey,
+    Consumer.TopicName (..),
+    Consumer.ConsumerRecord (..),
+    Partition.ProcessResult (..),
+    Consumer.PartitionId (..),
+    Consumer.unOffset,
+    Settings.Settings,
+    Settings.decoder,
+    -- For tests
+    processWithoutShutdownEnsurance,
```
ghost: Would be nice to move this into a private module.

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -0,0 +1,505 @@
...
+-- 3. A single-thread `pauseAndAnalyticsLoop` that tells the kafka library to pause & resume
+--    sending us messages from specific partitions (based on a number of factors)
+-- 4. Multiple worker-threads (`Partition.processMsgLoop`), one per
+--    (topic,partition)
+
+-- | Kafka.Worker is a module for processing a Kafka log.
+-- It can be used to build a CLI that will consume and process a user-defined message type
+module Kafka.Worker
+  ( process,
+    Description (..),
```
ghost: :art: I think we should not expose the constructor here. Otherwise adding a field to `Description` in the future will need to be a major version bump.

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -0,0 +1,505 @@
...
+    Partition.MessageCallback (Partition.MessageCallback),
+    OffsetSource (..),
+    Consumer.ConsumerGroupId (..),
+    PartitionKey,
+    Consumer.TopicName (..),
+    Consumer.ConsumerRecord (..),
+    Partition.ProcessResult (..),
+    Consumer.PartitionId (..),
```
ghost: :art: Lots of wrapper types here. I think it makes sense to have these inside the library, but wonder if we can tweak our API's so library users don't need to deal with them as much. Going to check.

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -0,0 +1,505 @@
...
+-- 4. Multiple worker-threads (`Partition.processMsgLoop`), one per
+--    (topic,partition)
+
+-- | Kafka.Worker is a module for processing a Kafka log.
+-- It can be used to build a CLI that will consume and process a user-defined message type
+module Kafka.Worker
+  ( process,
+    Description (..),
+    Partition.MessageCallback (Partition.MessageCallback),
+    OffsetSource (..),
```
ghost: Like `Description`, also think it'd be nice if we didn't expose the constructors for this one.

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -0,0 +1,505 @@
...
+  { partitions :: AllPartitions,
+    -- | When we receive a shutdown signal this variable will change. Threads
+    -- should stop processing new messages to allow the process to shut down.
+    stopping :: Stopping.Stopping,
+    analytics :: Analytics.Analytics,
+    rebalanceInfo :: RebalanceInfo
+  }
+
+-- | All the required parameters to start the Kafka worker
+data Description = Description
```
ghost: :art: Feels weird that we have both a `Settings` type and then a separate `Description` type. Can we combine these?

run-tests.sh
```diff
@@ -16,5 +16,38 @@ pg_ctl start -o '-k .'
...
+do
+  sleep 1
+done
+echo "kafka available"
+
+
 cabal build all
 cabal test all
+
+# cleanup
```
ghost, line 51: :bug: This will only run currently if the tests exit successfully. Usage will likely feel flaky.

ghost: Maybe this script has outgrown bash.

ghost (commented):
Looks great! I added a couple of notes, going to work through those.

nri-kafka/src/Kafka/Worker/Settings.hs
```diff
@@ -0,0 +1,150 @@
...
+        Environment.description = "Pausing reading from kafka when we have this many messages queued up but not yet processed",
+        Environment.defaultValue = "100"
+      }
+    (map MaxMsgsPerPartitionBufferedLocally Environment.int)
+
+decoderPollBatchSize :: Environment.Decoder Consumer.BatchSize
+decoderPollBatchSize =
+  Environment.variable
+    Environment.Variable
+      { Environment.name = "POLL_BATCH_SIZE",
```
ghost: :art: Missing `KAFKA_` prefix

nri-kafka/src/Kafka/Worker/Settings.hs
```diff
@@ -0,0 +1,150 @@
...
+        Environment.description = "The amount of messages we request in a single poll request to Kafka",
+        Environment.defaultValue = "100"
+      }
+    (map Consumer.BatchSize Environment.int)
+
+decoderOnProcessMessageFailure :: Environment.Decoder SkipOrNot
+decoderOnProcessMessageFailure =
+  Environment.variable
+    Environment.Variable
+      { Environment.name = "SKIP_ON_PROCESS_MESSAGE_FAILURE",
```
ghost: :art: Missing `KAFKA_` prefix

nri-kafka/src/Kafka/Worker/Partition.hs
```diff
@@ -0,0 +1,503 @@
...
+
+-- using this types as a proxy for named parameters
+-- this is a function that runs with the partition when it's done to cleanup
+newtype OnCleanup = OnCleanup (Prelude.IO ())
+
+-- | MessageCallback is used by your worker to process messages peeled off the queue.
+data MessageCallback where
+  MessageCallback ::
+    (Show e, Aeson.ToJSON msg, Aeson.FromJSON msg) =>
+    (Consumer.ConsumerRecord () () -> msg -> Task e ProcessResult) ->
```
ghost: :art: Might be nice to pass out a custom thing instead of `Consumer.ConsumerRecord () ()` here, so we can control documentation and don't require library users to use these kafka wrapper types.

run-tests.sh
ghost: I think we should change this, but not in this PR.

nri-kafka/src/Kafka/Worker/Internal.hs
```diff
@@ -46,11 +46,10 @@ data State = State
...
   }
 
--- | All the required parameters to start the Kafka worker
-data Description = Description
-  { settings :: Settings.Settings,
-    groupId :: Consumer.ConsumerGroupId,
-    topic :: Kafka.Topic,
+-- | The topics this worker should subscribe too. At the moment this library
+-- only supports subscribing too a single topic.
+data TopicSubscription = TopicSubscription
```
michaelglass, line 52: better name, too!

nri-kafka/src/Kafka/Worker/Settings.hs
```diff
@@ -148,3 +154,13 @@ decoderOnProcessMessageFailure =
...
         )
     )
+
+decoderGroupId :: Environment.Decoder Consumer.ConsumerGroupId
+decoderGroupId =
+  Environment.variable
+    Environment.Variable
+      { Environment.name = "KAFKA_GROUP_ID",
+        Environment.description = "Identifies the consumer group this worker is part off",
+        Environment.defaultValue = "procrastinators"
```
michaelglass: maybe we shouldn't have a default value at all?

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -22,7 +22,7 @@ module Kafka.Worker
     Internal.PartitionKey,
     Consumer.TopicName (..),
     Consumer.ConsumerRecord (..),
-    Partition.ProcessResult (..),
+    Partition.SeekCmd (..),
```
michaelglass: much better

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -15,18 +15,26 @@
 -- It can be used to build a CLI that will consume and process a user-defined message type
 module Kafka.Worker
   ( Internal.process,
-    Internal.TopicSubscription (..),
-    Partition.MessageCallback (Partition.MessageCallback),
+
+    -- * Settings
```
michaelglass, line 19: ✨

nri-kafka/src/Kafka/Worker/Internal.hs
```diff
@@ -54,6 +55,119 @@ data TopicSubscription = TopicSubscription
...
+-- >   settings <- Environment.decode decoder
+-- >   let subscription =
+-- >         subscription
+-- >           "the-topic"
+-- >           (\msg -> Debug.todo "Process your message here!")
+-- >   process settings subscription
+subscription ::
+  (Aeson.FromJSON msg, Aeson.ToJSON msg) =>
+  Text ->
+  (Envelope msg -> Task Text ()) ->
```
michaelglass: ✨

nri-kafka/src/Kafka/Worker/Internal.hs
```diff
@@ -203,7 +316,7 @@ rebalanceCallback skipOrNot observability callback offsetSource state consumer r
             log <- Platform.silentHandler
             fetchResult <- Task.attempt log (fetch newPartitions)
             case fetchResult of
-              Err err -> Exception.throwIO err
+              Err err -> Exception.throwString (Text.toList err)
```
michaelglass, line 303: why string?

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -26,19 +26,9 @@ module Kafka.Worker
     Internal.subscriptionManageOwnOffsets,
     Internal.Envelope (topicName, partitionId, messageOffset, payload),
     Partition.SeekCmd (..),
-
-    -- * Wrapper types
```
michaelglass: 🤘

nri-kafka/src/Kafka/Worker/Settings.hs
```diff
@@ -142,7 +142,7 @@ decoderOnProcessMessageFailure :: Environment.Decoder SkipOrNot
 decoderOnProcessMessageFailure =
   Environment.variable
     Environment.Variable
-      { Environment.name = "SKIP_ON_PROCESS_MESSAGE_FAILURE",
+      { Environment.name = "KAFKA_SKIP_ON_PROCESS_MESSAGE_FAILURE",
```
michaelglass, line 140: ❤️

nri-kafka/src/Kafka/Worker.hs
```diff
@@ -26,19 +26,9 @@ module Kafka.Worker
...
-    Internal.OffsetSource (..),
-    Consumer.ConsumerGroupId (..),
```
michaelglass: I believe we still need to export this now that it's an argument

nri-kafka/src/Kafka/Worker/Internal.hs
ghost: I changed `err` to be of type `Text`, which doesn't implement `Exception`, so we weren't able to use `throwIO` on it anymore.

nri-kafka/src/Kafka/Worker.hs
ghost: Maybe we should take it as a `Text` argument and wrap it ourselves.

nri-kafka/src/Kafka/Worker/Internal.hs
```diff
@@ -80,29 +80,15 @@ data Envelope msg = Envelope
 subscription ::
   (Aeson.FromJSON msg, Aeson.ToJSON msg) =>
   Text ->
-  (Envelope msg -> Task Text ()) ->
+  (msg -> Task Text ()) ->
```
ghost, line 79: Ooh, good point!

nri-kafka/src/Kafka/Worker/Internal.hs
```diff
@@ -113,29 +109,25 @@ subscriptionManageOwnOffsets ::
   (Aeson.FromJSON msg, Aeson.ToJSON msg) =>
   Text ->
   Task Text (Dict.Dict Int Int) ->
-  (Envelope msg -> Task Text Partition.SeekCmd) ->
+  (OffsetParams -> Int -> msg -> Task Text Partition.SeekCmd) ->
```
ghost: Liking the rename! Why is the offset a separate argument?

nri-kafka/src/Kafka/Worker/Internal.hs
```diff
@@ -108,10 +108,10 @@ subscription topic callback =
 subscriptionManageOwnOffsets ::
   (Aeson.FromJSON msg, Aeson.ToJSON msg) =>
   Text ->
-  Task Text (Dict.Dict Int Int) ->
+  ([OffsetParams] -> Task Text (Dict.Dict Int Int)) ->
```
ghost: This looks fine to me, but also I'm wondering if we're asking library users to do more work for a performance improvement that's probably not going to make much difference (considering the number of offsets will always be a lowish number).

nri-kafka/src/Kafka/Worker/Internal.hs
```diff
@@ -108,7 +108,7 @@ subscription topic callback =
 subscriptionManageOwnOffsets ::
   (Aeson.FromJSON msg, Aeson.ToJSON msg) =>
   Text ->
-  ([OffsetParams] -> Task Text (Dict.Dict Int Int)) ->
+  ([OffsetParams] -> Task Text (Dict.Dict OffsetParams Int)) ->
```
ghost: Much better!

ghost: Alternatively, we can add the offset back onto `OffsetParams` and return a list of them instead of a dictionary. Maybe call it `PartitionOffset`.

ghost: Did this!

ghost:
I got the current version compiling against our private repo.

I'm going to merge this so we can move forward using this in our private repo, but not publish to hackage yet in case there's additional changes we'd like to make.

nri-kafka/src/Kafka/Worker/Internal.hs
michaelglass: 🚨over-engineering🚨 perhaps the destination database will determine where to pull / put the data based on partition id?

agree we should drop it until it becomes an issue.

nri-kafka/src/Kafka/Worker/Internal.hs
```diff
@@ -102,14 +102,17 @@ subscription topic callback =
...
+-- >                "SELECT partition, offset FROM offsets WHERE partition = %"
+-- >                [partitions] )
 -- >           (\msg -> Debug.todo "Process your message here!")
 -- >   process settings subscription
 subscriptionManageOwnOffsets ::
   (Aeson.FromJSON msg, Aeson.ToJSON msg) =>
   Text ->
-  ([OffsetParams] -> Task Text (Dict.Dict OffsetParams Int)) ->
-  (OffsetParams -> Int -> msg -> Task Text Partition.SeekCmd) ->
+  ([Int] -> Task Text (List PartitionOffset)) ->
```
michaelglass, line 114: 👍

nri-kafka/src/Kafka.hs
```diff
@@ -144,8 +144,10 @@ topic :: Internal.Msg -> Text
 topic msg = Internal.unTopic (Internal.topic msg)
 
 -- | The payload of a message. This function might sometimes be useful in tests.
-payload :: Internal.Msg -> Maybe Aeson.Value
-payload msg = Maybe.map Aeson.toJSON (Internal.payload msg)
+payload :: (Aeson.FromJSON a) => Internal.Msg -> Maybe a
```
michaelglass, line 146: 👍

michaelglass:
thank you!

PR #7, opened 2021-07-01
Add show instance for Redis.Query based on Redis.Internal.cmds

stoeffel:


ghost (approved):
Would we want a `ToJSON` instance to? (assuming we're looking to use this in golden test results).

stoeffel:
> Would we want a `ToJSON` instance to? (assuming we're looking to use this in golden test results).

yes, makes sense. will do later today

stoeffel:
Actually, I think it's nicer to have the queries in the golden file with the output of show.

PR #9, opened 2021-07-07
Release script uploads docs too

ghost:
Without this hackage tries to generate docs, but this doesn't always work.


michaelglass:
merging this. Will revert if it causes other problems

PR #11, opened 2021-07-09
Release GHC 9.0.1 compatible versions

ghost:


ghost:
I have some concerns about `nri-redis` dependencies on `nri-test-encoding`. That transitively makes it depend on a bunch of `servant` libraries as well.

I wonder if we can cut the dependency. Some options:

- Let `nri-test-encoding` provide versions of the `nri-redis` api-creation functions that are `HasExamples`-enabled. Drop the `HasExamples` constraint in `nri-redis` itself.
- Extract the servant portions of nri-test-encoding into a separate lib (I like this one less, to be honest. Creates one more library, and keeps the dependency of `nri-redis` on `nri-test-encoding`).
- Change nri-redis to take an example value as an explicit argument, instead of using the HasExamples class. The user of nri-redis can still choose to use HasExamples to provide that value.
- ... ?

Thoughts?

stoeffel:
> Change nri-redis to take an example value as an explicit argument, instead of using the HasExamples class. The user of nri-redis can still choose to use HasExamples to provide that value.

I do think enforcing is better, but I can see how it's a problem.

> Extract the servant portions of nri-test-encoding into a separate lib (I like this one less, to be honest. Creates one more library, and keeps the dependency of nri-redis on nri-test-encoding).

don't like it either

> Change nri-redis to take an example value as an explicit argument, instead of using the HasExamples class. The user of nri-redis can still choose to use HasExamples to provide that value.

It's easy to forget adding a constraint though, but maybe that's the best option if we want to get rid of the dependency.

ghost:
> I do think enforcing is better, but I can see how it's a problem.

Yeah. Another thought I had with regards to this, is that while we do enforce adding an example, we don't enforce writing a test that uses the example. Assuming we don't forget the test (which currently we assume), then we can't forget adding the example either.

> It's easy to forget adding a constraint though

What do you mean by this? If the example value is passed in as an explicit argument we wouldn't be able to forget passing an example, right?

stoeffel:
> What do you mean by this? If the example value is passed in as an explicit argument we wouldn't be able to forget passing an example, right?

Oh, I think I understand what you mean now.

I wonder if an other option is to move the servant instances to webserver as orphans for now.

stoeffel:
do we need to do anything about this in this PR or how is it related to GHC 9.0.1?

ghost:
> do we need to do anything about this in this PR or how is it related to GHC 9.0.1?

It is related in two ways:
- The `HasExamples` constraint has not been released yet as a new version of `nri-redis`. We need to release a new version of nri-redis for supporting GHC 9.0.1. Including `HasExamples` would make that new version a major version bump that we otherwise don't need.
- It looks like the servant packages nri-redis depends on are currently marked as broken on stackage. We won't be able to reenable nri-redis on stackage after our GHC 9.0.1 version bump until those servant dependencies too are fixed.

ghost:
I made a commit changing the direction of dependency between `nri-redis` and `nri-test-encoding` for now. That would allow us to finish this task without making breaking changes now, so we can ponder the right solution a bit longer (and maybe do some other work first, per sync conversation this morning).

> I wonder if an other option is to move the servant instances to webserver as orphans for now.

That would work! Not much would remain in `nri-test-encoding` then. Should `nri-redis` implement its own example type?

I'm also starting to wonder if the initial premise is correct: we thought the `HasExamples` constraint would enforce some safety against API changes, but I now realize you don't get the safety without adding the test.

stoeffel:
> I'm also starting to wonder if the initial premise is correct: we thought the HasExamples constraint would enforce some safety against API changes, but I now realize you don't get the safety without adding the test.

true. I think it's good enough to do it on the "client" side.

PR #16, opened 2021-08-10
Print Kafka stop reason to stdout

ghost:
To help us figure out why we're shutting down in moments we're not
expecting it.


stoeffel (approved):
Looks good to me. thanks!

PR #18, opened 2021-08-25
Ignore ReceivedMsgNotInAssignedPartitions

stoeffel:


nri-kafka/src/Kafka/Worker/Fetcher.hs
```diff
@@ -11,7 +11,7 @@ import qualified Kafka.Worker.Partition as Partition
 import qualified Kafka.Worker.Settings as Settings
 import qualified Prelude
 
-type EnqueueRecord = (ConsumerRecord -> Prelude.IO Partition.SeekCmd)
+type EnqueueRecord = (ConsumerRecord -> Prelude.IO (Maybe Partition.SeekCmd))
```
ghost: 🎨 Instead of making this a `Maybe`, I propose we return a `Partition.NoSeek` instead of `Nothing`.

ghost (approved):
Proposed to get rid of the `Maybe`, otherwise this looks good to me!

nri-kafka/src/Kafka/Worker/Fetcher.hs
stoeffel: I wasn't sure if that would cause something else?

ghost:
Context on why I think this is good: the `throw` we added at a moment in development when we were a bit paranoid about accidentally skipping messages. We since started tracking kafka offsets in MySQL and that gave us trust we were not skipping messages. It makes me comfortable removing this `throw` too: in the unlikely/unexpected event it _does_ result in messages being skipped at the enqueueing step, we'd still catch it during processing.

PR #19, opened 2021-09-01
Scrollable span details

### LINKED ISSUE #17: log-explorer: make details scrollable

stoeffel:
fixes #17

### DIFF
--- a/nri-log-explorer/src/Main.hs
+++ b/nri-log-explorer/src/Main.hs
@@ -82,8 +82,11 @@ data SearchMatch = NoMatch | Matches Text
 data SpanBreakdownPageData = SpanBreakdownPageData
   { currentSpan :: RootSpan,
     search :: Search,
-    spans :: ListWidget.List Name (SearchMatch, Span)
+    spans :: ListWidget.List Name (SearchMatch, Span),
+    focus :: SpanBreakdownPageFocus
   }
+
+data SpanBreakdownPageFocus = FocusOnSpanList | FocusOnSpanDetails
 
 data FailureFilter = ShowAll | ShowOnlyFailures
 
@@ -226,11 +229,14 @@ update model msg =
                   Nothing -> page
                   Just (currentIndex, currentSpan) ->
                     SpanBreakdownPage
-                      SpanBreakdownPageData
-                        { currentSpan,
-                          spans = currentSpan |> logSpan |> toFlatList currentIndex,
-                          search = NoSearch
-                        }
+                      (
+                        SpanBreakdownPageData
+                          { currentSpan,
+                            spans = currentSpan |> logSpan |> toFlatList currentIndex,
+                            search = NoSearch,
+                            focus = FocusOnSpanList
+                          }
+                      )
         )
         |> andThen continueAfterUserInteraction
     Confirm ->
@@ -259,7 +265,8 @@ update model msg =
                           SpanBreakdownPageData
                             { currentSpan,
                               spans = currentSpan |> logSpan |> toFlatList currentIndex,
-                              search = NoSearch
+                              search = NoSearch,
+                              focus = FocusOnSpanList
                             }
         )
         |> andThen continueAfterUserInteraction
--- a/nri-log-explorer/src/Main.hs
+++ b/nri-log-explorer/src/Main.hs
@@ -893,7 +893,7 @@ viewContents page =
               viewSpanBreakdown spans
                 |> Brick.hLimitPercent 50
                 |> ( case focus of
-                       FocusOnSpanList ->
+                       FocusOnSpanDetails ->
                          Border.border
                        _ ->
                          identity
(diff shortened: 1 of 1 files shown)


nri-log-explorer/src/Main.hs
```diff
@@ -628,6 +630,7 @@ view model =
             viewContents page,
             viewKey page (clipboardCommand model)
           ]
+          |> Brick.joinBorders
```
stoeffel, line 700: This needs to be at the top for some reason.

nri-log-explorer/src/Main.hs
```diff
@@ -907,77 +913,81 @@ viewSpanBreakdownWithMatch matching span =
...
-            "source"
-            ( Data.Text.pack (Stack.srcLocFile srcLoc)
-                ++ ":"
-                ++ Text.fromInt (Prelude.fromIntegral (Stack.srcLocStartLine srcLoc))
-            ),
-      case Platform.succeeded original of
-        Platform.Succeeded -> viewDetail "result" "succeeded"
-        Platform.Failed -> viewDetail "result" "failed"
-        Platform.FailedWith exception ->
+  Brick.viewport
```
stoeffel, line 1014: I find it interesting that you added this inside `viewSpanDetails` where as I would have added it outside. I don't think it matters, but it's interesting how preferences are different.
I would have added it outside, because I feel like it's not part of rendering span details.

stoeffel: Also because you can write
```hs
viewSpanDetails span
  |> Brick.viewport SpanDetail Vertical
```
without having so much nesting inside here.

juanedi: > I find it interesting that you added this inside viewSpanDetails where as I would have added it outside. I don't think it matters, but it's interesting how preferences are different.

FWIW I also think that having it outside is better (reasoning: that feels more like "how it fits into the parent's layout"). I suppose that at the time I was so new to the code I didn't really understand what I was doing well enough to make that type of decision :-P

juanedi:
Status update after a brief pairing with @marcellerusu.

Done:
  - keeping track of currently focused pane and toggling with tab/backtab
  - ugly visual hint to be able to distinguish the currently focused pane (just went with whatever first occurred to test the logic)
  - keybinding hints at the bottom

To-dos that come to mind:
  - [x] bind scrolling to up/down and maybe j/k
  - [x] only scroll span details if we are focusing on that pane
  - [ ] visually pleasing hint of which is the current pane

juanedi:
I'll try to keep up the momentum and find the time to implement those this week.

stoeffel (commented):
Looks good so far!

juanedi:
Update: coloured borders are working but before merging I would still like to find a way to make the visual hint more clear. Since we always keep the center border highlighted, the only thing that changes is the side borders which are a bit hard to notice.

https://user-images.githubusercontent.com/753421/142001438-c35b4e47-6ebc-4bab-bd28-e74ff3f0f445.mov

stoeffel:
@juanedi I pushed a proposal for some focus styling. Code is ugly, but shows some ideas on how we can improve styling

juanedi:
Ohhhh I love it!!! It didn't occur to me that we could even do that 😅

nri-log-explorer/src/Main.hs
```diff
@@ -971,10 +985,13 @@ viewSpanBreakdown spans =
...
             |> if hasFocus
-              then Brick.withAttr "selected"
+              then
+                if focus == FocusOnSpanList
+                  then Brick.withAttr "selected"
+                  else Brick.withAttr "underlined"
               else identity
       )
-      True
+      (focus == FocusOnSpanList)
```
juanedi: It was hard for me to understand that `hasFocus` here refers to whether the current element is being focus, and that it was not referring also to whether the list widget was active. How would you feel about a rename in the lines of...:
- `hasFocus` -> `isSpanFocused`
- `focus` -> `focusedPane`

nri-log-explorer/src/Main.hs
```diff
@@ -1275,7 +1292,9 @@ attrMap =
               (204 :: Prelude.Integer)
               (102 :: Prelude.Integer)
           )
-      )
+      ),
+      ("dim-unfocused", Vty.withStyle Vty.defAttr Vty.Attributes.dim),
+      ("focused", Vty.withStyle Vty.defAttr Vty.Attributes.standout)
```
juanedi, line 1290: Cool! It didn't even occur to me that `Vty` would provide these styles. I guess I still don't know exactly what `Vty` is in charge of vs what falls under Brick's responsibilities 😅

nri-log-explorer/src/Main.hs
juanedi: Did that as part of https://github.com/NoRedInk/haskell-libraries/pull/19/commits/6402aac7b4397a3f47d20d917cc7147c42602b4c, along with other renames

juanedi:
🍐 'ed with @marcellerusu on some refactoring/renames. We didn't make any dramatic changes to the code. Mostly trying to consolidate all the conditional styling in one place.

@stoeffel this is ready for official review as far as we are concerned 🤘

stoeffel:
reviewed together with @juanedi ! thanks

PR #22, opened 2021-09-09
Zen/allow and check to return a value

stoeffel:
This allows an expectation chained with `andCheck` to return a value. This is useful in taskful tests.
See an example here https://github.com/NoRedInk/NoRedInk/pull/35216

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ ghcid-nri-observability-test: nri-observability/nri-observability.cabal
 nri-prelude/nri-prelude.cabal: nri-prelude/package.yaml
 	hpack nri-prelude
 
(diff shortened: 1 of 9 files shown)


ghost:
We had this before, but then changed it to what it was now in this commit:

https://github.com/NoRedInk/haskell-libraries/commit/2e50f6fabec6c115b4c854c06b7a98ab1fbf4095

tldr; is that if `andCheck` passes through a value it cannot be easily used as the last entry in a test, because the test itself expects a `Expectation' ()`, not any `Expectation' a`. I remember the error message you'd see in this case not being particularly clear either.

stoeffel:
oh you are right

PR #25, opened 2021-10-20
Update aeson to latest version

stoeffel:


stoeffel:
last error:

```

Data/Aeson/Encode/Pretty.hs:58:1: error:
    Could not find module ‘Data.Aeson.Encode’
    Perhaps you meant Data.Aeson.Encoding (from aeson-2.0.1.0)
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
58 | import qualified Data.Aeson.Encode as Aeson
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

builder for '/nix/store/j0hdcp9n072mw708y1idrajssqvkj4hp-aeson-pretty-0.8.8.drv' failed with exit code 1
cannot build derivation '/nix/store/7sq4fzwwl1la9kvrm1yp25pm9bhcfci4-ghc-8.10.4-with-packages.drv': 1 dependencies couldn't be built
error: build of '/nix/store/7sq4fzwwl1la9kvrm1yp25pm9bhcfci4-ghc-8.10.4-with-packages.drv' failed
```
going to try to overwrite aeson-pretty after lunch.

stoeffel:
next error is:
bugsnag-hs aeson mismatchhttps://github.com/jwoudenberg/bugsnag-hs/blob/trunk/package.yaml

ghost:
I updated it and there's a new version available on hackage. Looks like I closed the PR instead of merging it:
https://github.com/jwoudenberg/bugsnag-hs/pull/1

Will fix now.

jwoudenberg:
Wait, looks like I published a version 0.2.0.5 that increases the version bounds, and then a 0.2.0.6 that sets them back again. It's 0.2.0.5 we want. I deprecated 0.2.0.5.

ghost:
Wonder if we should drop this effort, pick it back up when other dependencies have had a bit more time to upgrade themselves.

stoeffel:
yes, I guess that makes sense:-(

PR #29, opened 2021-12-29
text-2.0.x base-4.16.x template-haskell-2.18.x [QUE-170]

omnibs:
Hackage poked us to relax version bounds

Note that these versions are untested because they're not in nixpkgs yet.

I did try adding text-2.0 through niv but it broke half the universe. Let's wait for unstable to catch up and move 9.0 there. Similarly, I tried adding a ghc921 build but everything was broken.

Hackage folks are on vacay and hackage registration emails are not firing, so I'm blocked from publishing for now.


stoeffel (approved):
LGTM! Thanks!

Issue #30, opened 2022-01-20
Raise bounds of hw-kafka-client to < 5.1 in order to allow version 5.0.0?


alaendle:
https://github.com/NoRedInk/haskell-libraries/blob/f9698d52078a6224cd88a99f1312ab024e51c2f0/nri-kafka/nri-kafka.cabal#L72

This would also help resolve https://github.com/commercialhaskell/stackage/issues/6388 (at least after a update nri-kafka was published).

PR #31, opened 2022-01-20
Raise hw kafka bounds

stoeffel:
fiixes #30


omnibs (approved):
Looks good. I wonder if we should revert #29 before merging this one. Merge this one, publish it, then create another PR for #29 where CI passes

PR #34, opened 2022-01-26
Set logfile mode only when necessary

charlieshanley:
Which should be the first time, when the user just created and owns the
file. Subsequently, the user running the code may not own the log file,
especially with multi-user nix installations, and thus won't be able to
change the file mode.


charlieshanley:
I tested this by running `niv update noredink-haskell-libraries -b logfile-mode-setting` in `.../NoRedInk/tooling`, then `firstaide-update`, then some log-generating commands (eg `aide watch rostering`), and reading and writing the log seemed to work fine despite the log file being owned by `_nixbld1`.

asterite (approved):
Looks great!

charlieshanley:
@NoRedInkBot r=asterite

PR #36, opened 2022-03-07
accept the slightly-broader Header' to allow for required headers

BrianHicks:
Right now we can't do

```haskell
Header' '[Required, Strict] "Key" Text
```

because we only implement `IsApi` for `Header`, which is an alias for `Header' '[Optional, Strict]`. We also can't add an orphan instance for it because `crawl` returns an opaque type. Loosening this up should still be valid for all instances of `Header`, however!

### DIFF
--- a/nri-test-encoding/src/Test/Encoding/Routes.hs
+++ b/nri-test-encoding/src/Test/Encoding/Routes.hs
@@ -25,7 +25,7 @@ import NriPrelude
 import qualified Servant
 import Servant.API
   ( Capture',
-    Header,
+    Header',
     QueryFlag,
     Raw,
     ReqBody',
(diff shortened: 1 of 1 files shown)


BrianHicks:
looks like this works where we need it, so I'm requesting review now

nri-test-encoding/src/Test/Encoding/Routes.hs
```diff
@@ -228,7 +228,7 @@ instance
     Examples.HasExamples val,
     IsApi a
   ) =>
-  IsApi (Header key val :> a)
```
Arkham, line 231: https://hackage.haskell.org/package/servant-0.19/docs/Servant-API-Header.html#t:Header-39-

nri-test-encoding/src/Test/Encoding/Routes.hs
```diff
@@ -228,7 +228,7 @@ instance
...
+  IsApi (Header' mods key val :> a)
```
Arkham, line 231: Servant exposed a `Header` which is `Header' '[Optional Strict]]` (from: https://hackage.haskell.org/package/servant-0.19/docs/Servant-API-Header.html#t:Header-39-)

Arkham (approved):
Thank you!

PR #38, opened 2022-03-28
Fix typo: outoing -> outgoing

waj:
This is a small typo but we should be aware that this will change the traces recorded in Honeycomb.


jali-clarke (approved):

jali-clarke:
@NoRedInkBot r+

PR #39, opened 2022-04-01
Allow starting a new root tracingSpan

zwilias:
This PR introduces `Platform.newRoot :: Stack.HasCallStack => Text -> Task e a -> Task e a`.

`newRoot` starts a new `tracingSpan` copying the behaviour of its root tracing span, so whatever `Task e a` it is given, is executed within a renamed copy of the root `LogHandler`. This could be used to start new "root" spans within a long lived process like a webserver handling requests, kafka interacting with things, or a loop executing a task at some interval eternally to ensure that things happening inside get handled as if they were happening at the root, with logging facilities properly writing out the interactions without waiting for the "real" root to actually finish executing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ ghcid-nri-observability-test: nri-observability/nri-observability.cabal
 nri-prelude/nri-prelude.cabal: nri-prelude/package.yaml
 	hpack nri-prelude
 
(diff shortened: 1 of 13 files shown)


stoeffel (approved):
Looks great!

PR #43, opened 2022-04-21
Cooperative stickies w/ metrics

omnibs:
Try out our fork of hw-kafka allowing cooperative rebalance strategy, based off of our WIP PR exposing rdkafka metrics, so we can understand what's up with rdkafka under the hood.

### DIFF
--- a/nri-kafka/src/Kafka.hs
+++ b/nri-kafka/src/Kafka.hs
@@ -159,7 +159,7 @@ key msg = Maybe.map Internal.unKey (Internal.key msg)
 handler :: Settings.Settings -> Maybe Stats.StatsCallback -> Conduit.Acquire Internal.Handler
 handler settings maybeStatsCallback = do
   producer <- Conduit.mkAcquire (mkProducer settings maybeStatsCallback) Producer.closeProducer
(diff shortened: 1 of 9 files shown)


nri-kafka/src/Kafka.hs
```diff
@@ -159,32 +159,32 @@ key msg = Maybe.map Internal.unKey (Internal.key msg)
...
 
 -- | We use a little trick here to poll events, by sending an empty message batch.
 -- This will call the internal pollEvent function in hw-kafka-client.
-pollEvents :: Producer.KafkaProducer -> Prelude.IO ()
-pollEvents producer = do
-  Producer.produceMessageBatch producer []
-    |> map (\_ -> ())
+flushProducer :: Producer.KafkaProducer -> Prelude.IO ()
+flushProducer producer = do
+  Producer.flushProducer producer
```
stoeffel, line 186: We've used `flushProducer` before in order to use sync sending messages. I think it's also a blocking function and I think to remember that it also slowed down sending immensely. ℹ️

stoeffel: maybe it's fixed though.

PR #45, opened 2022-07-20
Add task green thread ability

caiquefigueiredo:
This PR has two small changes:
- Make haskell-language-server work by using well-known ormolu workaround
- Allow Tasks to be run in a green thread. The Task result is then send to the provided callback

Worth noting: the same LogHandler will be used while executing the task.

### DIFF
--- a/nix/mk-shell.nix
+++ b/nix/mk-shell.nix
@@ -5,6 +5,12 @@ let
   workaround140774 = hpkg:
     with pkgs.haskell.lib;
     overrideCabal hpkg (drv: { enableSeparateBinOutput = false; });
+  # It is still necessary to run `hpack --force` into packages home dirs
+  haskell-language-server = pkgs.haskellPackages.haskell-language-server.override {
+    hls-ormolu-plugin = pkgs.haskellPackages.hls-ormolu-plugin.override {
+      ormolu = (workaround140774 pkgs.haskellPackages.ormolu);
+    };
+  };
 
 in pkgs.mkShell {
   buildInputs = [
@@ -59,17 +65,18 @@ in pkgs.mkShell {
         vector
         vty
       ]))
+    (workaround140774 pkgs.haskellPackages.ghcid)
+    (workaround140774 pkgs.haskellPackages.niv)
+    (workaround140774 pkgs.haskellPackages.ormolu)
     pkgs.apacheKafka # for nri-kafka
     pkgs.cabal-install
     pkgs.cachix
     pkgs.gnumake
-    (workaround140774 pkgs.haskellPackages.ghcid)
+    haskell-language-server
     pkgs.haskellPackages.hpack
-    (workaround140774 pkgs.haskellPackages.niv)
-    (workaround140774 pkgs.haskellPackages.ormolu)
     pkgs.pcre
-    pkgs.redis # for nri-redis
     pkgs.postgresql # for nri-postgres
+    pkgs.redis # for nri-redis
     pkgs.zlib
     pkgs.zookeeper # for nri-kafka
   ];
--- a/nri-prelude/src/Task.hs
+++ b/nri-prelude/src/Task.hs
@@ -47,7 +47,7 @@ import Platform.Internal (Task)
 import qualified Platform.Internal as Internal
 import Result (Result (..))
 import qualified System.Timeout
-import Prelude (IO, (*>))
+import Prelude (IO)
 import qualified Prelude
 
 -- BASICS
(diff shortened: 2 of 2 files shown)


nix/mk-shell.nix
```diff
@@ -59,17 +65,18 @@ in pkgs.mkShell {
...
       ]))
+    (workaround140774 pkgs.haskellPackages.ghcid)
+    (workaround140774 pkgs.haskellPackages.niv)
+    (workaround140774 pkgs.haskellPackages.ormolu)
     pkgs.apacheKafka # for nri-kafka
     pkgs.cabal-install
     pkgs.cachix
     pkgs.gnumake
-    (workaround140774 pkgs.haskellPackages.ghcid)
+    haskell-language-server
```
jali-clarke, line 75: should we include hls by default?  what if you don't use hls?

nri-prelude/src/Task.hs
```diff
@@ -3,6 +3,7 @@
 -- missing such a branch in a case statement looks like a problem and so is
 -- distracting.
 {-# OPTIONS_GHC -fno-warn-overlapping-patterns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
jali-clarke: it's not clear as to why we need this extension; we should be able to remove the type signatures in the let bindings within `background` and allow inference do its thing

nri-prelude/src/Task.hs
```diff
@@ -198,6 +200,23 @@ parallel tasks =
           |> Shortcut.map Prelude.sequence
     )
 
+-- | Given a task and a callback, execute the task in a greenthread
+-- and sends its result to callback.
+-- Always returns @OK ()@
+background :: forall x a . Task x a -> (Result x a -> Task x a) -> Task x ()
```
jali-clarke, line 205: would an additional variant that allows us to return a (wrapped) async value that we can then await on be useful?  feel free to veto if it's too low-level for our library

nri-prelude/src/Task.hs
```diff
@@ -198,6 +200,23 @@ parallel tasks =
...
+background task callback =
+  Internal.Task
+    ( \handler ->
+        let runTask :: Task x a -> IO (Result x a)
+            runTask task' = Internal._run task' handler
+            waitForTask :: Async.Async (Result x a) -> IO (Result x a)
+            waitForTask greenTask = do
+              result <- Async.wait greenTask
+              runTask <| callback result
+         in Async.withAsync (runTask task) waitForTask
```
jali-clarke: i believe `withAsync` blocks until the action is done, e.g.

```haskell
main = do
  result <-
    withAsync (threadDelay 5_000_000) $ \promise ->
      Async.wait promise
      pure 4
  print result
```

iirc it will wait 5s before printing `4`.  we may want to use the lower-level `async` instead

nix/mk-shell.nix
caiquefigueiredo: I did consider not pushing it. Have been holding it for a while.
Since it builds just fine, I feel like keeping it around, but I'd be completely fine if this were to be dropped should it become troublesome during an update or something like that.

nri-prelude/src/Task.hs
caiquefigueiredo: We really don't! Think I was using different type variables. Will change it

nri-prelude/src/Task.hs
caiquefigueiredo: I think that'd be really cool, but how would usage be? I like the current implementation because it keeps us within the Task monad with no user overhead/transformation/handlers.

nri-prelude/src/Task.hs
caiquefigueiredo: Indeed, but I think that happens due to the `result` dependency. It was my expectation that the usage of `*>` would avoid that block. Lemme check it :loading:

nri-prelude/src/Task.hs
jali-clarke: something like

```haskell
task :: Task x a
task = <...>

main = do
  asyncedTask <- ourAsync task
  result <- ourAwait asyncedTask
  doSomethingWith result
```

basically a `Task`-flavoured `async` / `await` that allows us to stay within `Task`.  this can be a bit error prone to use, though, for the reasons we should be using `withAsync` when dealing with `Async`s.  i do think we should _keep_ `background` as it is (modulo any other requested fixes); the `ourAsync` / `ourAwait` (or whatever) would exist beside `background`.  if it's not worth the effort, don't worry bout it

nix/mk-shell.nix
jali-clarke: (post-discussion) we're gonna leave it in.  shouldn't be too big a deal

nri-prelude/src/Task.hs
caiquefigueiredo: That is indeed the case 😮
Changed in d7944f648ef59d618872e05b0146013b2e25307f

nri-prelude/src/Task.hs
caiquefigueiredo: Very much like it, but would prefer to do it in another PR

nri-prelude/src/Task.hs
caiquefigueiredo: :curiosity: We actually do need it, cause otherwise won't be able to reconcile `background`s type variables with the ones we have inside the `let`. But it's already enabled as a default [here](https://github.com/NoRedInk/haskell-libraries/blob/1c996750f85816edbaee493cd92f6e22db94bb26/nri-prelude/package.yaml#L79)

jali-clarke: that's what i meant though - what happens if we remove the type declarations within the let?  should be ok, no?  (i realize that i'm bikeshedding here haha - please indulge me for just a little longer)

caiquefigueiredo: ooh, I see... probably, yeah
but I didn't want to leave it as inferred

jali-clarke (approved):

Ausqa21 (approved):
All good

PR #47, opened 2022-08-18
Allow for heterogenous expectation equality

caiquefigueiredo:
Add the `equalH` function so that we can have expectation between values of different types.


jali-clarke (approved):
lgtm, we paired on this.  reiterating concern that `equalH` should almost never _need_ to be used and that plain old `equal` should be used wherever possible

stoeffel:
Honestly, I'd prefer to keep edgecase functions like this where they are used instead of adding them to this module.

PR #51, opened 2022-08-30
Allow serialized test execution [QUE-350]

omnibs:
This PR allows us to specify groupings for test, which are then used by the test runner to serialize their execution.

The ideal future is to have these groupings:

```
[] -- no group
[mysql] -- mysql group
[postgres] --postgres group
[mysql, postgres] --group that uses both mysql and postgres
```

And have them execute like this:

```
parallel
  parallel
    []
  series
    [mysql]
  series
    [postgres]

series
  [mysql, postgres]
```

So we get maximum safety and maximum parallelism.

But right now, we're going for maximum safety:

```
parallel
  parallel
    []
  series
    [[mysql],[postgres],[mysql,postgres]]
```

### DIFF
--- a/nri-postgresql/src/Postgres/Test.hs
+++ b/nri-postgresql/src/Postgres/Test.hs
@@ -29,13 +29,15 @@ test ::
   (Postgres.Connection -> Expect.Expectation) ->
   Test.Test
 test description body =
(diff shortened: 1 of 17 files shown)


stoeffel (approved):
paired on review

PR #54, opened 2022-09-06
Fix groupping of tests [QUE-350]

stoeffel:


omnibs (approved):
Nice fix! CI is still unhappy, but the meat of it looks good

stoeffel:
> CI is still unhappy

Maybe it was those golden tests. I don't know why they changed. Removed the commit now.

PR #59, opened 2022-11-18
Add schema support to generatePGEnum

micahhahn:
Paired w/ Erik on this... we can support schema'ed enums slight better so let's do it.


erik-nri (approved):

micahhahn:
@erik-nri I borrowed the `[d|` splice thing from the `EnumSupport.hs` file which made the instances much more readable!  Also I was able to improve the array definitions so we don't need to required `UndecidableInstances` which seems like another clear win on top of the other better type errors.

PR #63, opened 2022-12-08
Add function to query keys

stoeffel:
This will be useful to clear out all `unsent-msg-*`

### DIFF
--- a/nri-redis/test/Main.hs
+++ b/nri-redis/test/Main.hs
@@ -261,7 +261,17 @@ queryTests redisHandler =
           |> Redis.query redisHandler
           |> Expect.succeeds
       field2
-        |> Expect.equal (Just "val2")
+        |> Expect.equal (Just "val2"),
+    Test.test "set and keys" <| \() -> do
+      Redis.set api "keys-1" "hello!" |> Redis.query testNS |> Expect.succeeds
+      Redis.set api "keys-2" "hello!" |> Redis.query testNS |> Expect.succeeds
+      result <- Redis.keys api "keys-*" |> Redis.query testNS |> Expect.succeeds
+      result
+        |> List.sort
+        |> Expect.equal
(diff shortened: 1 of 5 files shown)


nri-redis/test/Main.hs
```diff
@@ -261,7 +261,17 @@ queryTests redisHandler =
           |> Redis.query redisHandler
           |> Expect.succeeds
       field2
-        |> Expect.equal (Just "val2")
+        |> Expect.equal (Just "val2"),
+    Test.test "set and keys" <| \() -> do
+      Redis.set api "keys-1" "hello!" |> Redis.query testNS |> Expect.succeeds
+      Redis.set api "keys-2" "hello!" |> Redis.query testNS |> Expect.succeeds
```
juanedi, line 267: 🎨 adding a key that doesn't match would be nice to showcase the behavior of the function :-)

juanedi (commented):
Code looks great! HOWEVER...

It looks like our current issues with HQE are because we are completely failing to write anything to Kafka. The way I see it we shouldn't need to make aide clear out unsent messages from redis after we fix that. Am I missing anything?

If that is correct, I think I would prefer not exposing `keys` (or at least not without any warnings) as folks may be tempted to use it incorrectly. Quoting [the redis docs](https://redis.io/commands/keys/):

> Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout. Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using [SCAN](https://redis.io/commands/scan) or [sets](https://redis.io/topics/data-types#sets).

stoeffel:
yes, I think you are right!

stoeffel:
how annoying I can't close the PR on my phone

juanedi:
haha on it 💛

PR #69, opened 2023-01-23
Extract extending the expire of a query

stoeffel:


brian-carroll:
Once this is merged into haskell-libraries, how does it get into the monorepo? Is there some version bump that needs to happen? Seems like we should probably wait until `quiz-engine-http` is ready for the changes. I should be able to get that done in the next week or so.

brian-carroll:
Two related comments on this
- I would really like to change the name of `REDIS_DEFAULT_EXPIRY_SECONDS`. This is not just a default value for the `TTL` of a key. It is a much more elaborate behaviour.
- I think this whole expiry-extending behaviour should be moved out of `nri-redis` and into `quiz-engine-http`. That would make this library just an interface to Redis and the Quiz-Engine-specific behaviour would be in the Quiz Engine!

cc @zwilias

stoeffel:
> I think this whole expiry-extending behaviour should be moved out of nri-redis and into quiz-engine-http. That would make this library just an interface to Redis and the Quiz-Engine-specific behaviour would be in the Quiz Engine!

not addressed by this change? 

> I would really like to change the name of REDIS_DEFAULT_EXPIRY_SECONDS. This is not just a default value for the TTL of a key. It is a much more elaborate behaviour.

maybe a different PR?

> Once this is merged into haskell-libraries, how does it get into the monorepo? Is there some version bump that needs to happen? Seems like we should probably wait until quiz-engine-http is ready for the changes. I should be able to get that done in the next week or so.

we haven't published new versions to hackage in a while. We should change that, but for now we can follow this: https://github.com/NoRedInk/NoRedInk/blob/efac29e7a44a3171f5fa5fe78bdad0f685470e73/docs/how-tos/add-or-update-haskell-packages.md#L4

stoeffel:
@brian-carroll changed my mind. How about we change `extendExpire` to `extendExpire :: Int -> Query a -> Query a` and remove `REDIS_DEFAULT_EXPIRY_SECONDS` from nri-redis completely?

brian-carroll:
> @brian-carroll changed my mind. How about we change `extendExpire` to `extendExpire :: Int -> Query a -> Query a` and remove `REDIS_DEFAULT_EXPIRY_SECONDS` from nri-redis completely?

Nice idea!

The approach I was thinking of was to do something like `batch :: List (Query a) -> Query a` and then use that in HQE to make a helper that does an EXPIRE after every command.

Or maybe the existing `transaction` does that already? I'm not sure if a transaction has special meaning like it would in a relational DB.

So in that scenario `nri-redis` becomes just a wrapper around Redis' native functionality rather than building extra features around it. To me that is a simpler mental model of what `nri-redis` does. This expiry extension stuff is not built into Redis so I feel it doesn't belong here, even as an opt-in feature.

But your suggestions here are also good improvements!

stoeffel:
@brian-carroll I think you are right. I remove extendExpire and the setting completely. We can do this with a function in HQE like:
```hs
andExtendExpire : Redis.Api key a -> Int -> Redis.Query a -> Redis.Query a
andExtendExpire api secs q =
  Redis.map2 (\_ x -> x) (Redis.expire api secs) q
```

brian-carroll:
Great!
Changes look good. I read the doc about adding or updating these libraries. I think it might be best to wait to merge this PR until we're ready with the HQE changes. I plan to do that next week before Arthur starts to work on adding some more caches. (For scaffolding serializations and madlibs)

brian-carroll:
See also https://github.com/NoRedInk/NoRedInk/pull/42435

nri-redis/src/Redis/Internal.hs
```diff
@@ -182,19 +186,25 @@ sequence :: List (Query a) -> Query (List a)
 sequence =
   List.foldr (map2 (:)) (Pure [])
 
--- | The redis handler allows applications to run scoped IO
-data Handler = Handler
+data HasAutoExtendExpire = NoAutoExtendExpire | AutoExtendExpire
+
+data Handler' (x :: HasAutoExtendExpire) = Handler'
```
brian-carroll, line 196: what sorcery is this? 🧙 
So you can parameterize a type on a value?

brian-carroll: Maybe add some comments for us simple folk

brian-carroll (commented):
Nice solution!
I would just like to have some code comments explaining how we are using the more advanced language features, in case someone needs to make changes later who is less familiar with them.

brian-carroll (approved):
Great stuff! 👍

PR #70, opened 2023-02-14
Task background pool

stoeffel:
We'll need this for a feature in the main repo, in order to not explode background threads.

see https://github.com/NoRedInk/NoRedInk/pull/42572/files#diff-fcdccc3b7b18ef16a539a159cb2bffb67414a2ed600f0fe7a70d85404e30c063R100

### DIFF
--- a/nri-prelude/src/Task.hs
+++ b/nri-prelude/src/Task.hs
@@ -35,11 +35,13 @@ module Task
     concurrently,
     parallel,
     background,
+    backgroundPool,
   )
 where
 
 import Basics
 import qualified Control.Concurrent.Async as Async
+import qualified Control.Concurrent.Async.Pool as Async.Pool
 import qualified Internal.Shortcut as Shortcut
 import List (List)
 import qualified List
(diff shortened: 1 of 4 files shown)


stoeffel:
@zwilias do you understand why ci is failing?

zwilias:
Huh, weird, it's only failing on the GHC9.0 run. Does it work locally with the 9.0 shell.nix?

nri-prelude/src/Task.hs
```diff
@@ -222,6 +224,15 @@ background task =
         Prelude.pure <| Ok ()
     )
 
+backgroundPool :: Int -> Task Never a -> Task x ()
+backgroundPool poolSize task =
+  Internal.Task
+    ( \handler ->
+        Async.Pool.withTaskGroup (Prelude.fromIntegral poolSize) <| \g -> do
```
zwilias, line 231: If I'm reading thing correctly, the goal is to run this on one of `n` task pools, lazily creating them if they aren't there yet.

So, `backgroundPool 1 someTask` would add that task to "pool 1", `backgroundPool 2 someTask` would add that task to "pool 2", where the assumption is that each "pool" would roughly correspond to a thread, and the goal is to prevent resource starvation.

Assuming that understanding is correct:

it looks like `withTaskGroup` creates a new pool on every invocation and then adds a taskgroup with the specified size to that pool. So that's an unbounded growth on the number of pools, only limiting the amount of works that can happen in parallel within each taskgroup.

So I don't think those two match 🤔 Correct me if I'm wrong tho!

@jali-clarke might have some idea for how to divvy up work over a predefined number of workers, tho! At the very least, your input is welcome here 😄

stoeffel: oh, I actually think we probably want withTaskGroupIn

stoeffel: I'm going to fix this up properly tomorrow. On my phone now and with kids :-)

stoeffel: actually, maybe I'm just going to do this in place where it's needed, because I'm not sure what's a nice API yet.

stoeffel: Additionally, we've used `Data.Pool` in other places so I'm going to do that as well.

PR #73, opened 2023-02-24
Support testing conflicts in Postgres transactions

micahhahn:
In recent Phoenix work, we've made heavy use of database constraints to enforce consistency.  However, we're having trouble testing them!  Specifically `Postgres.Test.test` cannot recover from a transaction when a database error happens and gives this exception: 

```
PGERROR [25P02]: current transaction is aborted, commands ignored until end of transaction block
```

This is because at the end of `inTestTransaction` we are calling `rollbackAllSafe` ... which, counterintuitively, is not in fact safe.  `rollbackAllSafe` attempts to start a transaction with `pgBegin` to ensure there always is one, but this will fail if the transaction has already been aborted!

So the fix in this PR is to only call `rollbackAllSafe` when starting `inTestTransaction`, but make a normal call to `pgRollbackAll c` when wrapping up.  We should always have a transaction anyways when we reach the cleanup phase because the library doesn't expose a way to rollback the transaction to users.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,12 @@ nri-postgresql/nri-postgresql.cabal: nri-postgresql/package.yaml
 ghcid-nri-postgresql: nri-postgresql/nri-postgresql.cabal
 	cd nri-postgresql && ghcid
 
+setup-postgresql:
+	cd nri-postgresql && ./setup-postgres.sh
+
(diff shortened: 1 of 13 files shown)


stoeffel (approved):
Looks great! I pushed a few small commits. Do you mind reviewing those?

micahhahn (commented):
Thanks for the small tweaks! They look great!

PR #74, opened 2023-03-21
Change `Task.sequence` to actually run tasks in sequence 😨 

stoeffel:
This is a pretty fundamental change and there for a little scary, but I think it's the proper implementation of `Applicative` for `Task`.

## Current behavior

`Task.sequence` keeps on running tasks after a task has failed. 
This is that case because the `Applicative` instance of `Task` runs both parts before combining the `Result`s.

## Expected behavior

List of tasks given to `Task.sequence` runs until the first task fails or to completion.

## Fix

Run arguments to `(<*>)` in `Applicative` instance one after another.

Related internal discussion: https://noredink.slack.com/archives/C43KR2A5S/p1679431367295019

### DIFF
--- a/nri-prelude/src/Platform/Internal.hs
+++ b/nri-prelude/src/Platform/Internal.hs
@@ -68,18 +68,19 @@ instance Applicative (Task a) where
 
   (<*>) func task =
     Task <| \key ->
-      let onResult resultFunc resultTask =
-            case (resultFunc, resultTask) of
-              (Ok func_, Ok task_) ->
-                Ok (func_ task_)
-              (Err x, _) ->
-                Err x
-              (_, Err x) ->
-                Err x
-       in do
-            func_ <- _run func key
-            task_ <- _run task key
-            pure (onResult func_ task_)
+      do
+        func_ <- _run func key
+        case func_ of
+          Err x ->
+            Prelude.pure (Err x)
+          Ok okFunc ->
+            do
+              task_ <- _run task key
+              case task_ of
+                Err x ->
+                  Prelude.pure (Err x)
+                Ok okTask ->
+                  Prelude.pure (Ok (okFunc okTask))
 
 instance Monad (Task a) where
   task >>= func =
--- a/nri-prelude/src/Platform/Internal.hs
+++ b/nri-prelude/src/Platform/Internal.hs
@@ -10,6 +10,7 @@ import Control.Applicative ((<|>))
 import qualified Control.AutoUpdate as AutoUpdate
 import qualified Control.Concurrent.Async as Async
 import qualified Control.Exception.Safe as Exception
+import qualified Control.Monad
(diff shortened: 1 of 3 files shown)


stoeffel:
@jwoudenberg don't feel obligated to review, but I thought that this might be interesting for you :-)

nri-prelude/src/Platform/Internal.hs
zwilias, line 1: I think your analysis is correct: the core issue is that we're combining the `Result`'s after evaluating the internal `IO (Result x a)` functions.

Given that we have a seemingly lawful `Monad` implementation, however, we have alternatives! One would be a pretty straightforward implementation like so:
```
(<*>) func task = do
  func' <- func
  task' <- task
  Prelude.pure (func' task')
```

This is [the exact implementation of `Monad.Control.ap`](https://hackage.haskell.org/package/base-4.18.0.0/docs/src/GHC.Base.html#ap), however, so we could also go `(<*>) = Monad.Control.ap` and sidestep the whole ordeal.

zwilias: All tests pass with both implementations, so I'll leave it up to you. My personal preference would be to go with the very common approach of [`(<*>) = Monad.Control.ap`](https://github.com/search?q=%22%28%3C*%3E%29+%3D+ap%22+lang%3AHaskell+&type=code) rather than re-implementing that.

zwilias (approved):
Lovely!

jwoudenberg:
Don't have strong opinions! I guess for `sequence` the old behavior definitely doesn't match either the name `sequence` or the documentation for that function, so fixing that makes a lot of sense to me. Changing the applicative definition I think is trickier, as it affects `map2` ... `mapN` as well. On the plus side, sequential evaluation is how both Elm does it and how Haskell does it for `IO`, so there is precedent. On the downside, I think parallel evaluation might be the better default for something like `map2`, because it'll be faster. Or differently: if particular logic using `map2` requires sequential evaluation to function correctly then I think that logic would be more clearly written using `andThen` or do-notation. If that changes I think a careful review of existing applicative operations for `Task` would be good, because of the possibility of performance degredation.

stoeffel:
@jwoudenberg thanks for your input ❤️ 

>  If that changes I think a careful review of existing applicative operations for Task would be good, because of the possibility of performance degradation.

@jwoudenberg I think performance degradations are less of a worry to me. Let's look at an example:
```hs
(a, b) <- imChangingSomeStuffInRedis
  |> Task.map2 imChangingSomeStuffInRedisToo -- I'll fail
```
both tasks would be executed and leave some partial changes behind.
I know there are better ways of doing this, but I think this would be unexpected to me, and I'd expect that if one fails, the other will also fail.


EDIT after talking to @waj : There wasn't actually parallel evaluation, right? The second task was just executed after the first regardless of the first task's outcome.

waj (approved):
Works great!!

PR #76, opened 2023-04-11
Report Haskell tests run duration

charlieshanley:


nri-prelude/src/Test/Reporter/Stdout.hs
```diff
@@ -108,6 +120,8 @@ renderReport results =
                 failures,
             [ red (Text.Colour.underline "TEST RUN FAILED"),
               "\n\n",
+              black <| chunk <| "Duration:  " ++ Text.fromInt elapsed ++ " ms",
```
stoeffel: Should we report in seconds? I bet most tests take longer than a few ms.

nri-prelude/src/Test/Reporter/Stdout.hs
```diff
@@ -172,3 +186,20 @@ testFailure test =
           ++ "This is a bug.\n\n"
           ++ "If you have some time to report the bug it would be much appreciated!\n"
           ++ "You can do so here: https://github.com/NoRedInk/haskell-libraries/issues"
+
+elapsedMilliseconds :: List (Internal.SingleTest Platform.Internal.TracingSpan) -> List (Internal.SingleTest Internal.FailedSpan) -> Int
```
stoeffel: Very nit-picky, but I think this function does too much.
How about:
```hs
elapsedMilliseconds :: List TracingSpan -> Int

spans :: SuiteResult -> List TracingSpan

-- and maybe add something like
-- for rendering the elapsed time into a human readable format not only to ms
formatElapsedTime : Int -> Text
```

stoeffel (approved):
Nothing blocking a merge just some minor comments.
Thanks for adding this!

nri-prelude/src/Test/Reporter/Stdout.hs
asterite: I thought about this but the Elm test output shows it in `ms`. I guess seconds makes sense now that tests are taking about 5 seconds to run, though... I'll change it!

asterite:
@stoeffel I applied your suggestions, thanks! Let me know if it's good to go.

stoeffel (approved):
LGTM!

thanks

asterite:
@NoRedInkBot r+

PR #77, opened 2023-04-13
[nri-kafka] Interruption of attempt delay for graceful shutdown

waj:
Currently, when a Kafka worker is running and it fails to process a message, the task is retried with exponential backoff interval. At that moment, if the worker is signaled to stop (pressing ^C for example), the process actually doesn't stop until the attempt delay finishes, and actually it runs the task one more time.

This PR fixes that by moving the delay into the `peekRecord` function. That function already uses `Stopping.runUnlessStopping` that monitor the stop signal, effectively cancelling the delay.

Note that by moving the delay into that function it also solves the problem where the task is retried one more time before quitting.

### DIFF
--- a/nri-kafka/src/Kafka/Worker/Partition.hs
+++ b/nri-kafka/src/Kafka/Worker/Partition.hs
@@ -203,16 +203,6 @@ processMsgLoop skipOrNot messageFormat commitOffsets observabilityHandler state 
     StopThread ->
       Prelude.pure ()
     (NextMsg processAttempts record) -> do
-      case processAttempts of
-        (ProcessAttemptsCount 0) -> Prelude.pure ()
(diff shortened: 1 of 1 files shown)


stoeffel:
Is this a problem in production or just in development?

waj:
@stoeffel it could be a problem if the worker is failing to process messages. In such case it probably will not respond to graceful shutdown and the process might be killed by timeout.

stoeffel (approved):
LGTM!
It would be nice if we could add a test or something, but that might be hard

waj:
> It would be nice if we could add a test or something, but that might be hard

Yup, it's hard because it depends a lot on timing. Not imposible though. I'll try and send another PR if I succeed.

PR #78, opened 2023-04-14
[nri-kafka] Fix shutdown ensurance to actually do something

waj:
There's currently some code in `process` that should kill the process when a graceful shutdown is not possible. However that code is never executed when the process hangs during cleanup for example.

This PR fixes that by running the shutdown ensurance with a `race`. 

Also, it doesn't kill the process when the worker is stopped gracefully, making it possible to execute within a `ghci` session.


stoeffel (approved):
LGTM!

PR #80, opened 2023-04-25
Allow setting test timeout with NRI_TEST_TIMEOUT

zwilias:
On, e.g. CI, tests that take 2 or 3 seconds locally can easily take >10, just due to the sheer amount of things we're trying to do all at once.

Having generally fast tests is desirable, but - especially in internal/non-customer-facing services like the deployer that deal with `git` clones etc - not always feasible. We would most likely only set this to a higher number than 10s on CI, not locally.

### DIFF
--- a/nri-prelude/src/Test/Internal.hs
+++ b/nri-prelude/src/Test/Internal.hs
@@ -29,8 +29,10 @@ import Platform (TracingSpan)
 import qualified Platform
 import qualified Platform.Internal
 import qualified Set
+import qualified System.Environment
 import System.FilePath (FilePath)
 import qualified Task
+import Text.Read (readMaybe)
 import qualified Tuple
 import qualified Prelude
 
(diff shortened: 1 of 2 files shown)


nri-prelude/src/Test/Internal.hs
```diff
@@ -468,13 +470,22 @@ data Summary = Summary
...
   task'
     |> onException (Task.fail << ThrewException)
-    |> Task.timeout 10_000 TookTooLong
+    |> Task.timeout timeout TookTooLong
     |> Task.onError Task.fail
     |> Expectation
 
+timeoutFromEnvOrDefault :: Prelude.Double -> Expectation' Prelude.Double
+timeoutFromEnvOrDefault defaultTimeout = do
+  let fromIO io = Expectation <| Platform.Internal.Task (\_log -> map Ok io)
```
jali-clarke: let's move `Expect.fromIO` to be in `Test.Internal` and then re-export it from `Expect` since it imports `Test.Internal` anyway 😎

jali-clarke (approved):
seems legit

PR #81, opened 2023-04-26
[nri-http] Avoid buffer duplications

waj:
This PR moves a conversion of the response body to text, only to the branches interested in that result, reducing this way the required memory on some requests.

Also, this changes the type of the ByteString sent when using `expectBytesResponse` to use the lazy buffer returned by the `HTTP` library. This again avoids unnecessary duplication of buffers in memory making the evaluation of large responses more efficient. I bumped the version since this is a breaking change.
NOTE: even though the buffer is now a lazy ByteString, the response is fully evaluated by the time it comes out of the HTTP library (check [httpLBS](https://hackage.haskell.org/package/http-client-0.7.13.1/docs/Network-HTTP-Client.html#v:httpLbs)).

Closes ZEN-977

### DIFF
--- a/nri-http/CHANGELOG.md
+++ b/nri-http/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.4.0.0
+
+- Avoid unnecessary buffer duplication.
+- `expectBytesResponse` receives a lazy `ByteString`.
+
 # 0.3.0.0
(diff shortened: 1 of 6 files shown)


stoeffel (approved):
LGTM

PR #88, opened 2023-06-14
Fix delay calculation

waj:
This PR fixes an issue with the calculation of delay between Kafka message retries.
Before this, it was using the minimum between 1h and `10^attempts` seconds. But this later value (in microseconds) quickly exceeded the maximum value for an `Int64` and it was being truncated, resulting often either negative numbers or very small values, making the Kafka workers retrying extremely often and flooding BugSnag and Honeycomb with errors and traces.

The new formula caps directly the number of attempts before calculation and sets a maximum of `2^10` seconds (~17 minutes). I also changed the formula to use powers of 2 seconds to make a slower growth until the maximum.

Closes ZEN-991


omnibs (approved):
Nice catch!

PR #89, opened 2023-08-23
Add overload to specify golden result file

bcardiff:
The `Test.Encoding.Routes.test` function uses a fixed golden result file. This does not play nice if we want to split the routes in different services/areas.

This PR adds a `test'` variation to avoid the fixed file.


omnibs (approved):
Looks good!

PR #90, opened 2023-08-25
Don't bubble failure up

waj:
Currently, when a child of a tracing span is marked as failed, the parent is marked as failed as well. The problem with this is that it causes a lot of noise in Honeycomb and BugSnag for errors that are actually handled, like for example HTTP requests.

This PR removes that behavior and instead it adds another property to indicate that a trace contains at least one trace that has failed.


zwilias (approved):

omnibs (approved):
Approving with my changes.

I know my changes kinda defeat the purpose of this PR, since Waj wanted to sample out traces containing expected failed http requests in Rostering, but we'd also lose traces with Bugsnagged `Log.warn`, which we don't want.

Me and Waj have been talking about changes to `Http` that would allow the caller to decide whether a 404 or a timeout is a failure, for instance, which seem like a better way forward than sampling traces w/ errors.

PR #91, opened 2023-08-29
Define a Monad instance for Environment.Decoder

zwilias:
We've run into a case where we want the value of an environment variable to influence which other environment variables will be read. The most natural way of doing so is to use a monad instance.

For context, the particular case we're running into is that we want to read an environment variable defining which EKS cluster we're configured for, `and then` read the list of services that should be deployed on each environment that exists within that cluster.

### DIFF
--- a/nri-env-parser/src/Environment.hs
+++ b/nri-env-parser/src/Environment.hs
@@ -77,6 +77,7 @@ import Prelude
     IO,
     Integer,
     Integral,
+    Monad,
     Semigroup,
     fail,
     fromIntegral,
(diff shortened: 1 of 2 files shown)


nri-env-parser/src/Environment.hs
```diff
@@ -234,13 +235,27 @@ instance Applicative Decoder where
...
+        readFromEnvironment = \env ->
+          andThen
+            ( \val ->
+                toDecoder2 val
+                  |> readFromEnvironment
+                  |> \f' -> f' env
+            )
+            (f env)
+      }
+
```
jali-clarke, line 251: it's unfortunate that this isn't compatible with the `Applicative` instance semantics re: capturing both errors but there's really no choice there

jali-clarke (approved):
hmm introducing the ability to allow more complex logic during something simple like reading environment variables makes me a bit uneasy, though the alternative is multiple calls to `decode` and its compatriots, thereby typing the action as an `IO` instead of `Decoder`.  giving this a ✅

nri-env-parser/src/Environment.hs
zwilias: The second decoder only comes into existence if the first once succeeded, as opposed to the applicative semantics which can happen in parallel; so yeah, no choice indeed!

zwilias:
One alternative I considered was storing more complex data in a single environment variable, but that's pushing the boundaries of what "environment variables for configuration" should be, and not in a very positive direction.

Another alternative was not having the dependency on the cluster, and just reading all it could find, but again, that seemed less than ideal - configuration values that don't make sense in a given context, probably just shouldn't be read..

stoeffel:
Zero context here, but couldn't you do the same using https://hackage.haskell.org/package/nri-env-parser-0.1.0.8/docs/Environment.html#v:either and maybe adding a `oneOf` ( Like `JSON.Decode` in elm)?

PR #92, opened 2023-09-27
redis sorted sets

zwilias:
Add support for the [basic sorted sets commands](https://redis.io/docs/data-types/sorted-sets/#basic-commands). Further commands can be added later; hedis has pretty broad coverage!


waj (approved):
This looks great! Thank @zwilias!

PR #93, opened 2023-11-08
Add Redis.decoderWithCustomConnectionString

bcardiff:
Add a new decoder to grab the connection string from a specific connection string. 

We keep fetching values from REDIS_CLUSTER, REDIS_DEFAULT_EXPIRY_SECONDS, REDIS_QUERY_TIMEOUT_MILLISECONDS, REDIS_MAX_KEY_SIZE for the rest of the options.

### DIFF
--- a/nri-redis/src/Redis/Settings.hs
+++ b/nri-redis/src/Redis/Settings.hs
@@ -6,6 +6,7 @@ module Redis.Settings
     MaxKeySize (..),
     decoder,
     decoderWithEnvVarPrefix,
+    decoderWithCustomConnectionString,
   )
 where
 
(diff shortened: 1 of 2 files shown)


jali-clarke (commented):
could we add a comment in the code as to why we'd want to use the new decoder as opposed to the existing one?

nri-redis/src/Redis/Settings.hs
```diff
@@ -55,13 +56,24 @@ decoder :: Environment.Decoder Settings
 decoder =
   decoderWithEnvVarPrefix ""
 
+-- | decodes Settings from environmental variables with custom connection string
```
bcardiff, line 59: Would this be better @jali-clarke ?

```suggestion
-- | decodes Settings from environmental variables with specific name for the connection string variable
```

Other than that I find it hard to include something like, because we setup rails using an env var that does not follow our haskell services conventions and it's either this or adding a duplicated configuration. Which does not helps for public documentation 🙈

jali-clarke: yup all good - i'm all for adding this configurable, just wanted it to be clear as to why it exists in the first place

jali-clarke (approved):

PR #94, opened 2023-11-16
Add message offset to trace details for Kafka workers

omnibs:
When we log+trace kafka messages being assigned to workers, we don't log their offsets.

This makes it harder to inspect messages around it.

Inspecting preceding messages could have helped us root-cause Redis data loss in [PHX-1249](https://linear.app/noredink/issue/PHX-1249/finish-root-causing-hqe-redis-data-loss).


micahhahn (approved):
👌 

@NoRedInkBot r+

micahhahn:
@omnibs Feel free to merge this when CI passes

PR #103, opened 2024-03-21
Redis error can have eq

erik-nri:
While working through a refactor I found that existing tests that expected equality based on a Webserver Error would have be to modified to assert against a different Error type. That type I started to create had a variant that wrapped the `Redis.Error`. I found that `Redis.Error` did not have an `Eq` instance. And I thought to myself hmmm, why not?

This also removes the Haskell LS from nix. The version coming from nix was old, and did not work with ghc 9.4.7. I found that using ghcup to get the LS on version 2.5.0 built with 9.4.7 worked out great!

Lastly I also removed fplugin from `nri-redis` as we have in other places, and sorted out the imports as needed.

If this is cool, I am not sure what I need to do to go about bumping versions of things. Please let me know :).

### DIFF
--- a/nix/mk-shell.nix
+++ b/nix/mk-shell.nix
@@ -6,11 +6,6 @@ let
     with pkgs.haskell.lib;
     overrideCabal hpkg (drv: { enableSeparateBinOutput = false; });
   # It is still necessary to run `hpack --force` into packages home dirs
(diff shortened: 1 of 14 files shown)


erik-nri:
Closing for now. Did not end up needing the redis error thing. And other stuff is going on RN for the LS / nix stuff.

PR #104, opened 2024-04-01
add tshow

erik-nri:
We often end up using `Debug.toString` in places where we want to like... not actually debug something but just get a `Text`. It finally bugged me enough that I ran off to [Hoogle](https://hoogle.haskell.org/?hoogle=Show%20a%20%3D%3E%20a%20-%3E%20Text). From there I carefully ~stole~ ~borrowed~ literally copy pasted from the definition in RIO.Prelude.

### DIFF
--- a/nri-prelude/src/Text.hs
+++ b/nri-prelude/src/Text.hs
@@ -484,3 +484,10 @@ any = Data.Text.any
 -- > all isDigit "heart" == False
 all :: (Char -> Bool) -> Text -> Bool
 all = Data.Text.all
+
+-- | Get a Text representation of something Show-able.
+--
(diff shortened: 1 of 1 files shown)


stoeffel:
Historical context and than signing off again :-) In places where you don't want to debug a thing it's very likely better to turn it into a string explicitly using a custom xToString function instead of using this or Debug.toSteing. by being explicit you can actually turn it into something that is useful for the user and having control over how to do that.

erik-nri:
Thanks @stoeffel I appreciate the context and nice to see you again 😄.

celsobonutti (approved):

caiquefigueiredo (approved):

PR #106, opened 2024-04-19
Added atomic `putText` and `putTextLn` functions

waj:
This adds two functions that work like `putStr` and `putStrLn` but they use a lock so output doesn't get interleaved between multiple threads.

### DIFF
--- a/nri-prelude/src/Internal/IO.hs
+++ b/nri-prelude/src/Internal/IO.hs
@@ -15,11 +15,21 @@ import qualified Prelude
 stdoutLock :: MVar ()
 stdoutLock = unsafePerformIO (newMVar ())
 
+-- Output functions
+
+-- | Write text to stdout.
+--
+-- This function is similar to `Prelude.putStr` but receives a `Text` instead of a `Prelude.String`.
+-- Also this function is thread-safe, using a lock to prevent interleaved output.
 putText :: Text -> Prelude.IO ()
 putText text = do
   withMVar stdoutLock \_ ->
     Prelude.putStr (Text.toList text)
 
+-- | Write text to stdout followed by a newline.
+--
+-- This function is similar to `Prelude.putStrLn` but receives a `Text` instead of a `Prelude.String`.
+-- Also this function is thread-safe, using a lock to prevent interleaved output.
 putTextLn :: Text -> Prelude.IO ()
 putTextLn text = do
   withMVar stdoutLock \_ ->
--- a/nri-prelude/src/Internal/IO.hs
+++ b/nri-prelude/src/Internal/IO.hs
@@ -1,0 +1,26 @@
+{-# LANGUAGE BlockArguments #-}
+
+module Internal.IO
+  ( putText,
+    putTextLn,
+  )
+where
+
+import Control.Concurrent.MVar
+import System.IO.Unsafe (unsafePerformIO)
+import Text
+import qualified Prelude
+
+{-# NOINLINE stdoutLock #-}
+stdoutLock :: MVar ()
+stdoutLock = unsafePerformIO (newMVar ())
+
+putText :: Text -> Prelude.IO ()
+putText text = do
+  withMVar stdoutLock \_ ->
+    Prelude.putStr (Text.toList text)
+
+putTextLn :: Text -> Prelude.IO ()
+putTextLn text = do
+  withMVar stdoutLock \_ ->
+    Prelude.putStrLn (Text.toList text)
--- a/nri-kafka/scripts/pause-resume-bug/Producer.hs
+++ b/nri-kafka/scripts/pause-resume-bug/Producer.hs
@@ -7,7 +7,7 @@ import qualified Environment
 import qualified Kafka
 import Message
 import System.Environment (setEnv)
-import Prelude (IO, error, pure, putStrLn)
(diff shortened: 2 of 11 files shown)


stoeffel:
In case you don't know there is https://hackage.haskell.org/package/concurrent-output which might offer the same functionality.

waj:
@stoeffel I didn't know about that package, but if feels more complicated to achieve the same result. This PR is a simple fix for an issue that's currently so annoying.

nri-prelude/src/Internal/IO.hs
```diff
@@ -0,0 +1,26 @@
...
+{-# NOINLINE stdoutLock #-}
+stdoutLock :: MVar ()
+stdoutLock = unsafePerformIO (newMVar ())
+
+putText :: Text -> Prelude.IO ()
+putText text = do
+  withMVar stdoutLock \_ ->
+    Prelude.putStr (Text.toList text)
+
+putTextLn :: Text -> Prelude.IO ()
```
brian-carroll, line 33: If I was looking at a call site of this function, I would assume from the name that it was exactly the same as `putStrLn` except with `Text` instead of `String`. But in fact, if you use it, you are implicitly opting into some thread synchronisation stuff, and you may not realise it.

I think that is something that should be explicitly called out, perhaps by using a word like "lock" or "atomic" or something in the name.

Then if we ever somehow have a concurrency bug with this, it will at least "look suspicious".

nri-prelude/src/Internal/IO.hs
```diff
@@ -0,0 +1,26 @@
...
+  )
+where
+
+import Control.Concurrent.MVar
+import System.IO.Unsafe (unsafePerformIO)
+import Text
+import qualified Prelude
+
+{-# NOINLINE stdoutLock #-}
+stdoutLock :: MVar ()
```
brian-carroll, line 15: I just want to think through the safety of this.
The [docs for MVar](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html) say
> the "bigger" functions in this module ([swapMVar](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html#v:swapMVar), [withMVar](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html#v:withMVar), [modifyMVar_](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html#v:modifyMVar_) and [modifyMVar](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html#v:modifyMVar)) are simply the composition of a [takeMVar](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html#v:takeMVar) followed by a [putMVar](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html#v:putMVar) with exception safety. These have atomicity guarantees only if all other threads perform a [takeMVar](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html#v:takeMVar) before a [putMVar](https://hackage.haskell.org/package/base-4.19.1.0/docs/Control-Concurrent-MVar.html#v:putMVar) as well; otherwise, they may block.

We do meet those criteria, so that suggests that we do have atomicity guarantees. And I don't see how it could deadlock.
But I do think that we should have some code comments explaining why we think it's OK. Otherwise someone is going to come across this in a year and wonder.

Also I note that we do not have very many places where we print to stdout. It seems to be just on startup and shutdown of services. We do other things for logging.

Still, it is easy to make mistakes with this stuff, so it does make me feel a bit cautious. I wonder if it's worth having an env var to turn off the locking behaviour.

brian-carroll (commented):
Seems nice to have for local dev.
A few suggestions below.

nri-prelude/src/Internal/IO.hs
waj: Most languages have some sort of lock on file descriptors while printing strings. I don't know why Haskell decided not to do so (actually there's a lock on the fd, but not while accessing the buffer which ends in the same issue).
We could definitely include some comments to clarify that this function actually provides some guarantees on the output not being interleaved with concurrent calls. But I wouldn't make it explicit in the name, since that's what we all expect from those functions anyway.

nri-prelude/src/Internal/IO.hs
waj: I don't see any pattern here that could lead to a deadlock. The lock object is private to this module and only used within a `withMVar` block, which is exception-safe so it should never leave the lock unexpectedly locked.

waj:
@zwilias @brian-carroll do you see any reason to not merge this? It's a very simple solution to stop getting those scrambled useless outputs during development. It can be always changed to something better in the future.

zwilias:
I also wanted to think through deadlocking (in particular in things like `putTextLn =<< do putTextLn "foo"; pure "bar"`) but it seems safe, also with strict haskell.

Given these are exposed functions, I'd love to have haddock comments and an addition to the changelog, if you don't mind?

waj:
@zwilias done!

zwilias (approved):

PR #108, opened 2024-05-14
Update main nixpkgs to 23.11, drop GHC 8.10.7 support

zwilias:
- Update nixpkgs to release-23.11
- Make shell closure for ghc9.4 function again
- Drop support for GHC 8.10.7

Part of HACKY-830


jali-clarke:
no one should be using 8.10.7 at this point; anyone who is would be well-advised to use a previous version of our libs

ap-nri (approved):

PR #109, opened 2024-05-21
Add Lua Scripting support to nri-redis

omnibs:
Add an API to do Lua scripting to `Nri.Redis`.

We have a new Handler function `eval`, which does the recommended dance:
- Calculate the SHA1 of the Lua script we intend to run
- Try to run it with `EVALSHA ${sha1}`
  - On failure:
    - `LOAD SCRIPT ${theScript}`
    - Retry `EVALSHA ${sha1}`

The failure case only happens:
- The very first time a Lua script is tried on a specific Redis server
- The very first time any changes to the Lua script are tried on a specific Redis server

I initially tried to do this as a `Redis.Api` / `Internal.Query`, but it didn't work. If I understand semantics of type classes correctly, it won't unless `Internal.Query` is a `Monad`, not an `Applicative`.

Hedis uses a different `RedisResponse` typeclass for scripting results, which has its own decoding thingy. I wrote a few orphaned instances for common data types, like `Text`, `Int` and `()`.

Notes:
- The `eval` API supports namespacing of keys
- The `eval` API **does not** support auto-extending key expiry


nri-redis/test/Spec/Redis/Script.hs
```diff
@@ -0,0 +1,129 @@
...
+        |> Expect.equal "Script { luaScript = \"some text\", quasiQuotedString = \"some text\", keys = [], arguments = [] }",
+    Test.test "one key argument" <| \_ ->
+      [script|${Key "hi"}|]
+        |> printScript
+        |> Expect.equal "Script { luaScript = \"KEYS[1]\", quasiQuotedString = \"${Key \"hi\"}\", keys = [\"hi\"], arguments = [] }"
+        -- We can't test for compile-time errors, but manually test our helpful error message, uncomment
+        -- the lines below:
+        -- Test.test "compilation error" <| \_ ->
+        --   [script|${123}|]
+        --     |> Expect.equal "Doesn't matter, this won't compile"
```
micahhahn: Sooo, we can in fact test for compile-time errors :) 

The basic idea is that you call your function normally (not in a splice), and then there are tools in the `Language.Haskell.TH.TestUtils` module which let you run the `Q` monad and test the result. 

You might want to take a look at what I did in `nri-postgresql/test/Enum.hs` way back when I added postgres enum support.

micahhahn: Also, let's pair if this is something you would want to add!

nri-redis/test/Spec/Redis/Script.hs
```diff
@@ -0,0 +1,129 @@
...
+  ]
+
+mapLeft :: (a -> c) -> Either a b -> Either c b
+mapLeft f (Left a) = Left (f a)
+mapLeft _ (Right b) = Right b
```
micahhahn: 🎨 You could use [`mapFirst`](https://hackage.haskell.org/package/base-4.20.0.0/docs/Data-Bifunctor.html#v:first) instead from `Data.BIfunctor`

nri-redis/src/Redis/Script.hs
```diff
@@ -0,0 +1,286 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE IncoherentInstances #-}
```
micahhahn: I found a nicer way to avoid this: 

- #111

micahhahn (changes requested):
Amazing work! I really appreciate us going the extra mile here to make a good interface for this. 

A couple things to change below, but nothing deal breaking.

micahhahn (approved):

PR #110, opened 2024-05-28
migrate to GitHub actions for haskell-libraries ci

jali-clarke:
runs ci on github actions (incl. tests + push to cachix for shell closures) for macos + linux, and for each ghc version we support.  removes config for travis ci


fudomunro (approved):
Makes sense to me. I suppose there might be things that fail in Github actions that weren't failing in Travis, but I can't imagine anything critical.

PR #111, opened 2024-05-28
Patch to avoid using `IncoherentInstances`

micahhahn:
My editor was flagging `IncoherentInstances` in red, presumably because it is a deprecated language extension. 

image

I did a bit of [reading on overlapping instances](https://serokell.io/blog/learn-from-errors-overlapping-instances#overlapping-instances-with-type-variables) and found that we can just specify `INCOHERENT` directly on the instance and avoid the language extension!


omnibs (approved):
Oh, awesome, thank you for linking the article and improving comments!

PR #112, opened 2024-06-03
Fix ordering of `script` argument interpolation

omnibs:
I wrote a `foldl` that builds up a list of keys and literals for `nri-redis`' `script` quasi quote, and forgot to unrevert the lists 🤦 

This PR fixes it.


micahhahn (approved):
😬 

You know what, I swear I thought about this myself when reviewing the last PR and somehow convinced myself that it was working. 

FWIW, I usually would use names like `revKeys` or `revArguments` for these kinds of folds to remind myself to reverse at the end. I've also had thoughts about making a separate `RevList` type that you can only concat to and would reverse on converting to a `List`.

omnibs:
> I've also had thoughts about making a separate RevList type that you can only concat to and would reverse on converting to a List.

Omg. Brilliant.

PR #114, opened 2024-10-01
nri-http: fix timeout value in microseconds

waj:
According nri-http docs, the `timeout` value should be specified in microseconds. However this value is being multiplied by 1000 before passed to `http-client` library which also expects the value to be in microseconds.

Effectively we already use this value throughout our codebase as microseconds, so it makes sense to fix the code and not the docs.


michaelglass:
@waj @zwilias & I simplified CI. If you rebase on trunk it should pass now

zwilias (approved):

PR #116, opened 2024-10-17
add redis sismember

michaelglass:
adds sismember to Redis set api


bcardiff (approved):
I kicked the CI again, the failure seems unrelated to this change.

PR #117, opened 2024-10-21
get CI working & misc cleanups

michaelglass:
- drops testing of old GHC versions because they were breaking CI / not building
- drop deprecated github actions (and upgrade nix installer while we're at it)
- upgrade CI nixpkgs to match what's in niv
- drop deprecated `Data.Pool.createPool`
- run tests script stops postgres (but not redis)

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Documentation comments use the [Haddock][haddock] format to ensure they are rend
 
 ### Testing compilation
 
-Our default `shell.nix` sets up an environment around the most recent version of GHC that we support. CI also runs tests against a couple of older versions of GHC. To test using any of these versions locally:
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -32,7 +32,7 @@ Documentation comments use the [Haddock][haddock] format to ensure they are rend
 
 ### Testing compilation
 
-Our default `shell.nix` sets up an environment around the most recent version of GHC that we support. CI also runs tests against a couple of older versions of GHC. To test using any of these versions locally:
+Our default `shell.nix` sets up an environment around the most recent version of GHC that we support. CI used to supporst running tests against a different versions of GHC. To test using any of these versions locally:
```
zwilias:
```suggestion
Our default `shell.nix` sets up an environment around the most recent version of GHC that we support. CI used to support running tests against a different versions of GHC. To test using any of these versions locally:
```

zwilias (changes requested):
> supporst

zwilias (approved):

PR #123, opened 2024-11-27
better observability for redis timeouts

michaelglass:
this PR ensures redis failures include trace information

previously, it was discarded by `Task.timeout`

PR #124, opened 2024-11-27
can modify redis timeout mid stream

michaelglass:
adds the ability to modify query timeout on a handler

### DIFF
--- a/nri-redis/CHANGELOG.md
+++ b/nri-redis/CHANGELOG.md
@@ -1,4 +1,10 @@
+# 0.2.0.3 unreleased
+
+- [bugfix] When a query times out, its context is no longer removed from the Stack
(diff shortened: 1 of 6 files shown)


nri-redis/CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
+# 0.2.0.3 unreleased
+
+- [bugfix] When a query times out, its context is no longer removed from the Stack
```
michaelglass, line 3: this change is in #123

bcardiff (approved):
Having such flexibility makes sense. But I worry it could end up being hard to track in application code.

michaelglass:
yeah I think ... maybe this should have a tracing span

PR #125, opened 2024-11-28
add a tracing span when modifying the query timeout

michaelglass:
follow-up to #124. This makes it easier to see in logs if the lib consumer has changed the timeout settings.

### DIFF
--- a/nri-redis/test/Spec/Redis.hs
+++ b/nri-redis/test/Spec/Redis.hs
@@ -23,7 +23,7 @@ import qualified Prelude
 -- put this at the top of the file so that adding tests doesn't push
 -- the line number of the source location of this file down, which would
 -- change golden test results
-spanForTask :: (Show e) => Task e () -> Expect.Expectation' Platform.TracingSpan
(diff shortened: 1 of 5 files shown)


nri-redis/test/Spec/Redis.hs
```diff
@@ -134,15 +134,30 @@ observabilityTests handler =
...
+          handlerThatExpiresImmediately <- Expect.succeeds (Redis.withQueryTimeoutMilliseconds 0 handler)
           span <-
-            Redis.query (Redis.withQueryTimeoutMilliseconds 0 handler) (Redis.ping api)
+            Redis.query handlerThatExpiresImmediately (Redis.ping api)
               |> spanForFailingTask
           span
             |> Debug.toString
             |> Expect.all
-              [ Expect.equalToContentsOf (goldenResultsDir ++ "/observability-spec-timeout-reporting-redis-query"),
+              [ Expect.equalToContentsOf (goldenResultsDir ++ "/observability-spec-reporting-redis-query-timeout"),
```
michaelglass, line 144: following naming convention of the rest of the tests

bcardiff (approved):

PR #129, opened 2025-02-05
nri-kafka: expose retry counts (for worker managing its own offsets)

omnibs:
We're exposing how many times a message has been retried to consumer subscription handlers.

This allows consumers to implement a dead letter queue.

We're exposing that only for consumers that manage their own offsets, because that's where we're building a DLQ rn. If more people want DLQs, we can do it for regular `subscribe` too.

### DIFF
--- a/nri-kafka/test/Main.hs
+++ b/nri-kafka/test/Main.hs
@@ -9,7 +9,7 @@ import qualified Prelude
 main :: Prelude.IO ()
 main = do
   -- macos runners seem to be slow and fail on several kafka integration tests
-  System.Environment.setEnv "NRI_TEST_TIMEOUT" "20000"
+  System.Environment.setEnv "NRI_TEST_TIMEOUT" "240000"
   Test.run tests
 
 tests :: Test.Test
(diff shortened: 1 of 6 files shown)


omnibs:
@brian-carroll friendly ping, in case this slipped through, bc the GH plugin doesn't ping the team for non-monorepo PRs

brian-carroll:
> @brian-carroll friendly ping, in case this slipped through, bc the GH plugin doesn't ping the team for non-monorepo PRs

Oops, yeah totally missed it, sorry!

nri-kafka/test/Main.hs
```diff
@@ -2,11 +2,15 @@ module Main (main) where
...
 import qualified Spec.Kafka.Worker.Partition
+import qualified System.Environment
 import qualified Test
 import qualified Prelude
 
 main :: Prelude.IO ()
-main = Test.run tests
+main = do
+  -- macos runners seem to be slow and fail on several kafka integration tests
+  System.Environment.setEnv "NRI_TEST_TIMEOUT" "240000"
```
brian-carroll, line 12: Wow this is a lot.
I thought it was just a temporary thing while you were debugging but I see it's still there.
I guess if we need it, we need it!
How long do the tests actually take?

brian-carroll (approved):
I'm not sure if the test timeout is a bit excessive but am approving to unblock

PR #132, opened 2025-04-24
Add `deriving Generic` to `Result` (for aeson & elm-export)

brian-carroll:
Add `deriving Generic` on `Result`.
I want to generate an Elm decoder for a response that contains a `Result` but I can't! And I can't add the right instances in my app code! I am amazed we never ran into this before.

I wondered about other types but it turns out that `Result` is the only sum type that we actually _define_ in `nri-prelude`. `Maybe` and `List` are just aliases for types defined in the standard prelude, and `Dict` is an alias for `Map`. That's why all of those have `Generic` instances.

Part of QUO-1404

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,11 @@ jobs:
         uses: easimon/maximize-build-space@v10
         with:
           # you can try to increase this if the ubuntu runner runs out of space.
-          # prob can't push it too far above 30GB though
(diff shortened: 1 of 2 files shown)


omnibs (approved):
I'm surprised you couldn't create an [orphan instance](https://wiki.haskell.org/index.php?title=Orphan_instance)!

Regardless, a non-orphan one is preferable

PR #133, opened 2025-05-16
nri-observability: async reports

waj:
Reports are queued and executed on a separate thread. Wait for the queue to be empty when the handler is released.

This is useful for short lived processes like cron jobs, where a single trace is reported. Currently the report is executed using `async` without waiting it to finish, so the traces are lost.

Part of ZEN-2516

### DIFF
--- a/nri-observability/src/Observability.hs
+++ b/nri-observability/src/Observability.hs
@@ -20,7 +20,10 @@ module Observability
 where
 
 import qualified Conduit
+import Control.Concurrent.Async (async, wait)
+import Control.Concurrent.STM.TQueue (newTQueueIO, readTQueue, writeTQueue)
 import qualified Control.Exception.Safe as Exception
+import Control.Monad.STM (atomically)
 import qualified Data.Aeson as Aeson
 import qualified Environment
 import qualified List
@@ -43,6 +46,10 @@ newtype Handler = Handler
   }
   deriving (Prelude.Semigroup, Prelude.Monoid)
 
+data Msg
+  = Report Text Platform.TracingSpan
+  | Close
+
 -- | Function for creating an observability handler. The settings we pass in
 -- determine which platforms we'll report information to.
 handler :: Settings -> Conduit.Acquire Handler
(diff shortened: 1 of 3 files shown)


omnibs:
Asked Brian to review and assess the interplay of async reporting and Haskell-in-Ruby, bc I'm in a bit of a rush rn

Context is when control is given back to ruby after a servant-in-rails request, haskell has no chance to run async code anymore till it gets its next call

brian-carroll (changes requested):
Yeah as Juliano mentioned, I think we have to test that this doesn't break Haskell in Ruby monitoring. Threads only make progress when HiR calls are being made.
I'm going to mark this as "request changes" until we figure it out.

A few possible ways forward:
- Test if this works correctly with HiR. Maybe the problems we saw before with HTTP will be different here for some reason. I am not sure we have a test setup that can do this though. Maybe review environment could help.
- Somehow detect that we are in HiR and go back to sync reporting in that case (perhaps using some monolith env var?)
- Provide sync and the async versions of the handler or make the handler into a record with two fields. I'm not sure what impact this has on other code.

waj:
Thanks @brian-carroll! Having separate functions to do sync and async reporting sounds like a good alternative.
I wonder how's this working today, since the report call made from `rootTracingSpanIO` is always wrapped in an `async` call.

brian-carroll:
Ah I didn't realise it was already async. From the description I thought async was new. I reviewed the diff and didn't see any other async calls. But now I realise that of course it must be somewhere that didn't change! I will try to find it.

As far as I understand, the GHC runtime advances the async tasks a little bit whenever it is running. But when control flow returns to Ruby, _all_ execution stops within GHC runtime. That means all the async tasks are paused until it starts executing again on the next FFI call, which could be from a different HTTP request from a different user.

I guess if you have a steady stream of FFI calls happening then everything should eventually get done.
There is a latency that is not ideal but apparently it's not new on this PR.
I can't remember exactly what problem this caused in the past, I will look back on that.

So what difference could this PR make? 🤔 

The questions I'm asking myself now:
With this PR are we generating a larger number of async tasks per FFI call?
How many async tasks can GHC get done per FFI call?
If each FFI call generates 5 new tasks but can only get 3 older tasks completed, are they piling up indefinitely?
I need to research more about how this actually works.

brian-carroll:
OK I finally managed to dig up the right context on this from Dec 2023
https://github.com/NoRedInk/NoRedInk/pull/46644
Luckily I wrote a detailed PR description.
In that PR we switched from the single-threaded runtime to the multi-threaded runtime to fix Honeycomb monitoring.
@omnibs and I were both remembering the details of a problem that we _fixed_ in that PR.
Since that time we have been running with multithreaded GHC RTS and the HTTP requests to Honeycomb have been working just fine.
Our observability module is already using async, this PR just changes how that works.
So everything is OK here
@waj sorry for the confusion on this.

brian-carroll (dismissed):

waj:
I'm pausing the implementation of this PR for now, because just enqueuing all requests makes the reporting serialized across all backends. Making this right would require have individual queues for each backend and doing some batching, etc...

For now I'm opening #134 to do something similar to the current behavior, but making the release of the observability handler to wait for pending reports.

PR #134, opened 2025-05-23
nri-observability: flush reports when handler is being closed.

waj:
- Move asynchronicity from `Platform.rootTracingSpanIO` to `Handler.report`.
- Observability handler waits for pending reports while releasing.

This is useful for short lived processes like cron jobs, where a single trace is reported. Currently the report is executed using async without waiting it to finish, so the traces are lost.

Part of ZEN-2516


brian-carroll (approved):
Looks good.
Initially I was afraid the allocation limit stuff might have a lot of edge cases to think about, but then I realised that code is just moved from somewhere else, which is reassuring!

PR #135, opened 2025-05-27
Wait for the test root span to be reported

waj:
This fixes an issue that started when #134 was merged. Some tests started to fail because they were spawning other traces with `Platform.newRoot` and that was sending multiple spans to the same reporter. Previously those reports were executed within an `async` but now the report is done synchronously and it failed to put the second trace into the `MVar`.


jali-clarke (approved):
thank you!

PR #136, opened 2025-06-11
Add Test.runWithSettings

waj:
This adds a `Test.runWithSettings` function that allows customizing where to write the output (console, JUnit file and NRI logs). This is useful to write specs on the output without having to fork the process and capture the console, which is unsafe when multithreading.


bcardiff (approved):

waj:
Closing this for now. Even though it might be a good idea, it didn't help fixing the issues I was originally working on.

PR #137, opened 2025-07-13
Add NamedRoutes support to nri-test-encoding

bcardiff:
Also bumps min servant version to match the currently used one. This avoids the need for conditional compilation.


waj:
LGTM. Do you want to open a PR in our main repo pointing to this version? So we can verify that it can pass CI before merging.

bcardiff:
Based on a local branch the code should work.

The lower bound versions I got them from monorepo dev environment.

The code is not yet used, but i can check things will build with the version bump if that is where the uncertainty is coming.

waj (approved):

PR #138, opened 2025-08-01
support GHC 9.10.2, GHC 9.12.2; drop GHC 9.4.x

jali-clarke:
also bump nixpkgs to `release-25.05`.  with love and regards to hello@perk.ee

todo:
* [x] tweak the branch protection rule to pick up the ghc-9.12 ci job as a required check (we don't use the macos job as a required check since it sometimes times out or runs out of space)

completes HACKY-1187


perkee:
ILYSM. It isn't even quitting time and you got it done. Goodness.

derek-noredink (approved):
Would you steal open source to feed your family?

perkee:
I'm sure I can make make a package where I get the source from here rather than hackage/stackage, but please also publish pls they're so hungry.

PR #140, opened 2025-08-20
`expectBytes`: mismatches on `Http.Request' errorType` are ... awkward

omnibs:
(see commit message comments)

### DIFF
--- a/nri-http/src/Http/Mock.hs
+++ b/nri-http/src/Http/Mock.hs
@@ -161,7 +161,7 @@ tryRespond [] req =
       handleCustomResponse :: (Internal.Response s -> Result e expect) -> Task e (a, expect)
       handleCustomResponse f = case f (Internal.NetworkError_ msg) of
         Err err -> Task.fail err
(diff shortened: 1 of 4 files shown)


nri-http/src/Http/Mock.hs
```diff
@@ -161,7 +161,7 @@ tryRespond [] req =
       handleCustomResponse :: (Internal.Response s -> Result e expect) -> Task e (a, expect)
       handleCustomResponse f = case f (Internal.NetworkError_ msg) of
         Err err -> Task.fail err
-        Ok _ -> Debug.todo "Since we manually craft the Response as an Error, this case will not run."
+        Ok _ -> Debug.todo msg
```
brian-carroll, line 164: Developer: "This thing will not happen"...
Real world: Thing happens
😅

omnibs: what's super weird is this was not happening on your branch in the monorepo

you were getting a message saying "no stub for `GoogleWhatevsResponse`" instead of hitting the `Debug.todo`

here in haskell-libraries i was hitting the `Debug.todo`

PR #141, opened 2025-09-22
Fix space leak in nri-observability

omnibs:
Since #134 was deployed to prod, NRI's highest traffic service started slowly increasing memory usage over time, at a rate of around 5MiB/min.

This PR introduces a test app that reproduces the memory leak, and a fix.

The issue was elusive:
- Profiling with `-hm` pointed fingers at `Data.Conduit.Internal`
  - [This blogpost](https://www.well-typed.com/blog/2016/09/sharing-conduit/) made me pretty certain it was the new conduit, but I don't fully follow the blogpost, in all honesty
  - `-fno-full-laziness` didn't help
- [ndmitchell's recipe for finding space leaks](https://ndmitchell.com/downloads/slides-plugging_space_leaks_improving_performance-06_oct_2016.pdf) didn't work
  - Using `-K1K` didn't make my leaky test app blow up
- Profiling with `-hy` showed an `Int` was leaking, and Waj's PR did introduce an Int inside the `STM` in the handler

I was still suspicious of `Conduit`, so while googling for `conduit stm space leak` I landed on [the `strict-stm` library](https://hackage.haskell.org/package/strict-stm), which claimed to have helped make `cardano-node` leak-free.

The library was a drop-in replacement for `stm`, and using it made our leak go away.

|before|after|
|---|---|
|image|image|

### DIFF
--- a/nix/common-haskell-overrides.nix
+++ b/nix/common-haskell-overrides.nix
@@ -36,4 +36,8 @@ self: super:
 
     # for now, pin hw-kafka-client to 4.0.3; nixpkgs@release-25.05 provides 5.3.0
     hw-kafka-client = self.callHackage "hw-kafka-client" "4.0.3" { };
+
+    # marked broken in nixpkgs
+    # current version requires io-classes >=1.5 && <1.6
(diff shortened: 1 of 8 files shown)


brian-carroll (approved):
Excellent! Great detective work! 🕵️ 🔍

PR #143, opened 2025-09-23
Lighterweight `Platform.silentHandler`

omnibs:
While digging for space leaks in Nri's Amazonka wrapper, I noticed a bunch of `TracingSpan` and `TracingSpanDetails` building up where a `Platform.silentHandler` was being used.

Digging into it, I noticed `silentHandler` uses `mkHandler`, which creates lots of `IORefs` which, in `silentHandler`'s case, were never useful.

I made an internal `nullHandler` function specifically for the public `silentHandler` that does no IO.

I still kept `silentHandler : IO LogHandler`, even tho I could have made it pure, to keep it a non-breaking change, because I was lazy.

### DIFF
--- a/nri-prelude/src/Platform/Internal.hs
+++ b/nri-prelude/src/Platform/Internal.hs
@@ -616,6 +616,21 @@ mkHandler requestId clock onFinish onFinishRoot' name' = do
         finishTracingSpan = finalizeTracingSpan clock allocationCounterStartVal tracingSpanRef >> andThen onFinish
       }
 
+-- | Helper that creates a handler that does nothing. This is intended to power
+-- basically @Platform.silentHandler@ and nothing else. We provide this to make
+-- @Platform.silentHandler@ as efficient as possible, skipping all side effects.
+nullHandler :: LogHandler
+nullHandler = do
+  LogHandler
+    { requestId = "",
+      startChildTracingSpan = \_ -> pure nullHandler,
+      startNewRoot = \_ -> pure nullHandler,
+      setTracingSpanDetailsIO = \_ -> pure (),
+      setTracingSpanSummaryIO = \_ -> pure (),
+      markTracingSpanFailedIO = pure (),
+      finishTracingSpan = \_ -> pure ()
+    }
+
 -- | Set the details for a tracingSpan created using the @tracingSpan@
 -- function. Like @tracingSpan@ this is intended for use in writing libraries
 -- that define custom types of effects, such as database queries or http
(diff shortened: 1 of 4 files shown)


nri-prelude/src/Platform/Internal.hs
```diff
@@ -616,6 +616,21 @@ mkHandler requestId clock onFinish onFinishRoot' name' = do
         finishTracingSpan = finalizeTracingSpan clock allocationCounterStartVal tracingSpanRef >> andThen onFinish
       }
 
+-- | Helper that creates a handler that does nothing. This is intended to power
+-- basically @Platform.silentHandler@ and nothing else. We provide this to make
+-- @Platform.silentHandler@ as efficient as possible, skipping all side effects.
```
brian-carroll, line 621: It would be nice to mention memory leaks specifically.
Engineers often disagree on the importance of efficiency, whether certain special cases are justified, etc.
But a memory leak is pretty objective. And if someone changes this code, they need to watch out for that.

brian-carroll (approved):

nri-prelude/src/Platform/Internal.hs
omnibs: Ah, I didn't mention memory leaks because I'm not really sure the leaks are caused by `mkHandler`. We use `mkHandler` all the time with other `LogHandler` instances. What's special about `silentHandler`? Remember we didn't see leakage in high-throughput apps like HQE before #134? Every request in HQE uses `mkHandler`.

I think it might be more of an issue of how the handler is being used, than the handler itself.

There's [this old article by Edsko de Vries](https://www.well-typed.com/blog/2016/09/sharing-conduit/) on Conduit and space leaks, which I don't really follow, that reinforces the suspicious there's something about our usage that makes this leak. It seems it's not exactly what Edsko describes tho, because `-fno-full-laziness` didn't help.

This is also why I was more keen on removing the unused `silentHandler` logger in our Amazonka wrapper, than fixing it to actually log things to the request's tracing span. If I knew what's really causing the leak and how to fix it, I'd rather have logs.

omnibs: Although I could leave a word of warning wrt this whole situation. Someone might be grateful.

brian-carroll (approved):
Perfect, thanks!

nri-prelude/src/Platform/Internal.hs
brian-carroll: Yeah exactly 👍

PR #147, opened 2025-11-21
nri-http: Global settings for user agent and default timeout

waj:
This adds a `Http.handlerWith` function that accepts `HttpSettings` to configure a default timeout for all requests and automatically adds a `User-Agent` header. By default, the agent is set to `nri-http/<version>`, but it can be overridden via the settings.

It also includes a fix to prevent duplicate `X-Request-ID` headers from being sent with requests.


adauguet (approved):
Looks good! Go Canvas!

PR #148, opened 2025-11-25
Make ormolu aware of infix operators

waj:
Add a `.ormolu` file to have the right indentation of infix operators.

### DIFF
--- a/nri-postgresql/src/Postgres/Settings.hs
+++ b/nri-postgresql/src/Postgres/Settings.hs
@@ -320,9 +320,9 @@ toPGDatabase
             then
               Text.toList host
                 </> ".s.PGSQL."
-                  ++ show port
-                  |> SockAddrUnix
-                  |> Right
+                ++ show port
+                |> SockAddrUnix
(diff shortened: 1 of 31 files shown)


nri-postgresql/src/Postgres/Settings.hs
```diff
@@ -320,9 +320,9 @@ toPGDatabase
             then
               Text.toList host
                 </> ".s.PGSQL."
-                  ++ show port
-                  |> SockAddrUnix
-                  |> Right
+                ++ show port
+                |> SockAddrUnix
+                |> Right
```
micahhahn, line 325: This makes my brain hurt a bit, but I assume the operator precedence is such that this can be safely read left to right.

micahhahn (approved):
This looks much more readable! Thanks for the fix!

nri-postgresql/src/Postgres/Settings.hs
waj: Yup, I agree!

PR #149, opened 2026-01-29
Add `Expect.runExpectation`

celsobonutti:
This adds a new function `Expect.runExpectation`,  which takes an Expectation and runs it into an `IO`. This is necessary for using the `Llm.Test.Cached` API from the `llm-client` library we now have.

### DIFF
--- a/nri-prelude/src/Expect.hs
+++ b/nri-prelude/src/Expect.hs
@@ -59,6 +59,7 @@ module Expect
     Internal.runExpectation,
     Internal.fromIOResult,
     Internal.Expectation',
+    Internal.Failure,
     around,
   )
 where
--- a/nri-prelude/src/Expect.hs
+++ b/nri-prelude/src/Expect.hs
@@ -57,6 +57,7 @@ module Expect
     -- * Fancy Expectations
     Internal.fromIO,
     Internal.runExpectation,
+    Internal.fromIOResult,
     Internal.Expectation',
     around,
   )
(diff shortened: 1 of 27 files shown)


nri-prelude/src/Test/Internal.hs
```diff
@@ -394,6 +394,20 @@ fromIO io =
   Platform.Internal.Task (\_ -> map Ok io)
     |> Expectation
 
+-- | Run an expectation directly in IO.
+-- On success, returns the result. On failure, throws the Failure as an exception.
+runExpectation :: Expectation' a -> Prelude.IO a
+runExpectation expectation = do
+  log <- Platform.silentHandler
```
micahhahn: I'm a bit worried about using `silentHandler` here.  That will just not log anything that we do inside of our test setup which is probably not what we want. 

Let's modify this function to take `LogHandler` as an argument and then use that logger for Task.perform.

Inside of `withMocks` we should be able to call `Expect.succeeds Platform.logHandler` to get the handler to use with `runExpectation`.

micahhahn (changes requested):

micahhahn (approved):
Looks good! LGTM

micahhahn (changes requested):
Actually hold on a second... this exception test wasn't exactly what I was thinking

nri-prelude/src/Expect.hs
```diff
@@ -56,6 +56,8 @@ module Expect
 
     -- * Fancy Expectations
     Internal.fromIO,
+    Internal.runExpectation,
+    Internal.fromIOResult,
     Internal.Expectation',
     around,
```
micahhahn, line 63: I believe we will also need to expose `Failure` right?

micahhahn (changes requested):

micahhahn (approved):
Alright, now we are looking good to merge

PR #151, opened 2026-05-07
Improve sendsync performance

omnibs:
# Context
Our synchronous writes to Kafka have been slow since forever, and we used to think that was on Kafka itself: high throughput but high latency.

Turns out it was a simple mistake on our side.

Synchronous sends are synchronous as much as Erlang/Elixir's synchronous messaging is synchronous: it's a wrapper around an async send that synchronously waits on an async ack.

When a kafka broker acks a message, librdkafka's I/O thread enqueues a delivery-report event onto an internal queue. That event sits there until some thread calls `rd_kafka_poll`, which drains the queue and dispatches callbacks. Those callbacks free up pending `sendSync` calls.

# The issue

In our codebase, `pollEvents` is what triggers `rd_kafka_poll`, and we used to have it like this:

```
pollEvents producer = do
  Producer.produceMessageBatch producer []
    |> map (\_ -> ())
  Control.Concurrent.threadDelay 100_000 {- 100ms -}
  pollEvents producer
```

That 100ms delay means `sendSync` might take up to 100ms to be waken up after an ack comes back from the broker.

# Motivation

Having fast writes to Kafka allows us to use `syncSend` and forego complex async retries and delivery confirmations in a new service.

# The fix

We could have simply reduced the delay, but there's another function called `flushProducer`, which does something similar to what we did, but in a better way.

`flushProducer` hangs on a call to `rd_kafka_poll` for up to 100ms. If no events come, it times out, and you can call it again. If an event comes, it return immediately.

The 100ms timeout might seem superfluous, but ceding control back to Haskell allows us to respond to async exceptions and terminate our producers cleanly.

`hw-kafka`'s own [example producer](https://github.com/haskell-works/hw-kafka-client/blob/main/example/ProducerExample.hs#L94) uses this function.

## Benchmarks

We created a simple benchmark script here to test sync sends with and without the fix.

```sh
$ KAFKA_BROKER_ADDRESSES=localhost:9092 BENCHMARK_MESSAGE_COUNT=200 BENCHMARK_WARMUP=20 cabal run sync-write-benchmark -fsync-write-benchmark
```

### Without the fix

```
Benchmarking sync writes: warmup=20 count=200 topic=nri-kafka-sync-benchmark
Warming up...
Measuring 200 sends...
count=200  min=100.1ms  avg=101.1ms  p50=101.1ms  p95=101.4ms  p99=102.1ms  max=102.1ms
```

### With the fix

```
Benchmarking sync writes: warmup=20 count=200 topic=nri-kafka-sync-benchmark
Warming up...
Measuring 200 sends...
count=200  min=5.6ms  avg=6.3ms  p50=6.2ms  p95=6.9ms  p99=8.6ms  max=8.8ms
```

This opens the possibility to adopt synchronous Kafka writes more broadly.

# New issue

We uncovered a new issue while digging into this: we're not handling delivery failures in `sendSync`.

We just wrote a `known-issues.md` documenting future work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         runner:
-          - os: macos-14 # macos sonoma on m1; 3 vcpu, 7 GB memory
+          - os: macos-26 # macos tahoe on m1; 3 vcpu, 7 GB memory
             base-nixpkgs-channel: nixpkgs-25.05-darwin
           - os: ubuntu-22.04 # linux x86_64; 4 vcpu, 16 GB memory
             base-nixpkgs-channel: nixos-25.05
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,33 +24,19 @@ jobs:
           - ghc-9-12
     runs-on: ${{ matrix.runner.os }}
     steps:
-      # see write-up @ https://github.com/marketplace/actions/maximize-build-disk-space
+      # GitHub-hosted ubuntu runners don't have enough free disk for the nix
+      # store; this removes preinstalled software to make room.
       - name: Reclaim space
         if: matrix.runner.os == 'ubuntu-22.04'
-        uses: easimon/maximize-build-space@v10
+        uses: jlumbroso/free-disk-space@v1.3.1
         with:
-          # you can try to increase this if the ubuntu runner runs out of space.
-          # prob can't push it too far above 30GB though.  the github actions runners
-          # have a small disk size, and if we want more we'll have to look into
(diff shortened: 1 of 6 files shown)


omnibs:
looping in Waj too in case he's less busy. i _think_ he has also used nri-kafka

i'm testing these changes in backyard rn and won't merge before i finish testing

omnibs:
Just tested this in backyard.

I fired 100 `HistorySaveChangedUsages` commands to HQE in quick succession. I ran it from `aide rails-console backyard`.

We got 8ms average duration:
image

Compared to staging:
image

brian-carroll (approved):

brian-carroll:
Great speedup, well done!

PR #155, opened 2026-05-28
Bump libs after nri prelude 0.7

omnibs:
I bumped nri-prelude to 0.7 in #152 bc of breaking changes.

I should have also bumped these packages, whose version boundaries against nri-prelude changed due to the breaking change.

I'm planning to publish them all so I bumped the copyright year too (our release.sh script checks it).


jali-clarke (approved):
pretty straightforward!

PR #159, opened 2026-06-13
nri-kafka: fix sendSync hanging forever on delivery failure

omnibs:
## Problem

*Note: We never actually observed the problematic behavior in production or anywhere. It might be we never failed to write a synchronous message to Kafka 😅. Claude pointed out this issue during the `sendSync` performance work in #151, and since we'll rely on `sendSync` in the event platform, I thought we'd better have this covered.*

`Kafka.sendSync` blocked on a `TMVar` that the delivery callback only signaled on the *success* branch. When a message was enqueued but the broker ultimately failed to deliver it (`delivery.timeout.ms` exceeded, retries exhausted, a non-retriable broker error, or no available partition leader), the `TMVar` was never written and the calling request handler **parked forever** — and the failure was muted from observability.

The synchronous `ImmediateError` enqueue-refused path already failed correctly; this was specifically the asynchronous, post-enqueue delivery-report path.

## Fix

- Add a pure `Internal.deliveryReportToResult :: DeliveryReport -> Result Error ()` that maps **every** delivery report to a result, and signal the terminator on all branches.
- A failed delivery now surfaces as a descriptive `Task.fail`:
  - `DeliveryFailed (ProducerRecord, KafkaError)` — broker-side delivery failure (carries the record).
  - `NoMessageDelivered KafkaError` — the message-less `NoMessageError` report (librdkafka's delivery callback fired with a null message pointer; error read from `errno`).
- `sendHelperAsync`'s callback now receives the whole `DeliveryReport` and fires on every report (was success-only). `sendAsync` keeps its prior **success-only** callback contract.
- The exactly-once-signal contract is preserved (librdkafka invokes the delivery callback exactly once per message).

`Error` lives in the unexposed `Kafka.Internal` module, so the new constructors are not a public-API change.

## Testing

- New pure unit tests in `test/Spec/Kafka.hs` cover all three report branches (developed test-first: watched them fail against a stub, then pass).
- Full worker **integration suite passes against a live broker** (12/12), confirming the rewiring doesn't regress `sendSync`'s success path.
- The broker-side delivery-*failure* path has no automated integration test by design — a deterministic delivery failure is inherently flaky to reproduce, which is why the dispatch logic was extracted into the pure, unit-tested function.

## Docs

- `docs/known-issues.md` entry marked resolved (original write-up kept for context).
- `CHANGELOG.md` entry added under a new `# Unreleased` heading.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/nri-kafka/src/Kafka.hs
+++ b/nri-kafka/src/Kafka.hs
@@ -215,18 +215,32 @@ mkHandler settings producer = do
           Platform.tracingSpan "Async send Kafka messages" <| do
             let details = Details (List.map Producer.unBrokerAddress (Settings.brokerAddresses settings)) msg'
             Platform.setTracingSpanDetails details
-            sendHelperAsync producer doAnything onDeliveryCallback msg'
+            -- Preserve the existing async contract: the caller's callback only
+            -- fires once the broker has confirmed delivery. Failures are not
+            -- forwarded to async callers (they would have to be reported out of
+            -- band), only to sync callers below.
+            let onDeliveryReport deliveryReport =
+                  case deliveryReport of
+                    Producer.DeliverySuccess _producerRecord _offset -> onDeliveryCallback
+                    _ -> Task.succeed ()
+            sendHelperAsync producer doAnything onDeliveryReport msg'
               |> Task.mapError Internal.errorToText,
         Internal.sendSync = \msg' ->
           Platform.tracingSpan "Sync send Kafka messages" <| do
             let details = Details (List.map Producer.unBrokerAddress (Settings.brokerAddresses settings)) msg'
             Platform.setTracingSpanDetails details
             terminator <- doSTM doAnything TMVar.newEmptyTMVar
-            let onDeliveryCallback = doSTM doAnything (TMVar.putTMVar terminator Terminate)
-            sendHelperAsync producer doAnything onDeliveryCallback msg'
+            -- The callback runs on every delivery report, so the terminator is
+            -- signalled exactly once whether delivery succeeds or fails. This
+            -- is what keeps a failed delivery from parking the caller forever.
+            let onDeliveryReport deliveryReport =
+                  doSTM doAnything (TMVar.putTMVar terminator (Internal.deliveryReportToResult deliveryReport))
+            sendHelperAsync producer doAnything onDeliveryReport msg'
               |> Task.mapError Internal.errorToText
-            Terminate <- doSTM doAnything (TMVar.readTMVar terminator)
-            Task.succeed ()
+            result <- doSTM doAnything (TMVar.readTMVar terminator)
+            case result of
+              Ok () -> Task.succeed ()
+              Err err -> Task.fail (Internal.errorToText err)
       }
 
 doSTM :: Platform.DoAnythingHandler -> STM.STM a -> Task e a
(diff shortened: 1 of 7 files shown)


nri-kafka/src/Kafka/Internal.hs
```diff
@@ -48,13 +48,44 @@ instance Show Encodable where
...
+    -- delivery callback fires with a null message pointer and the error code is
+    -- read from @errno@. It is an exceptional, library-level condition rather
+    -- than a normal per-message broker rejection (those arrive as
+    -- 'DeliveryFailed', which carries the record).
+    NoMessageDelivered Producer.KafkaError
   | Uncaught Exception.SomeException
   deriving (Show)
 
 errorToText :: Error -> Text
 errorToText err = Text.fromList (Prelude.show err)
```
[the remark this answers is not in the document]
omnibs: This was already the case for `SendingFailed`. I think this is better addressed by ensuring our messages wrap PII in e.g. Log.Secret, so they're still debuggable, than omitting their contents completely.

nri-kafka/test/Spec/Kafka.hs
```diff
@@ -0,0 +1,45 @@
...
+              Err (Internal.NoMessageDelivered kafkaError) ->
+                Expect.equal kafkaError exampleError
+              other -> Expect.fail ("expected Err (NoMessageDelivered ...), got " ++ Debug.toString other)
+        ]
+    ]
+
+exampleRecord :: Producer.ProducerRecord
+exampleRecord =
+  Producer.ProducerRecord
+    { Producer.prTopic = "the-topic",
```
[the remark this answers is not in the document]
omnibs: Fixed

nri-kafka/src/Kafka.hs
```diff
@@ -215,18 +215,32 @@ mkHandler settings producer = do
...
-            sendHelperAsync producer doAnything onDeliveryCallback msg'
+            -- Preserve the existing async contract: the caller's callback only
+            -- fires once the broker has confirmed delivery. Failures are not
+            -- forwarded to async callers (they would have to be reported out of
+            -- band), only to sync callers below.
+            let onDeliveryReport deliveryReport =
+                  case deliveryReport of
+                    Producer.DeliverySuccess _producerRecord _offset -> onDeliveryCallback
+                    _ -> Task.succeed ()
+            sendHelperAsync producer doAnything onDeliveryReport msg'
```
omnibs, line 226: Note that this is unchanged from `trunk`. We do nothing on failure in the async path.

Fixing this is a tad larger than what we're aiming for here, and I'm not 100% sure how it would work. I've tried and failed to have a `LogHandler` report stuff in code that runs in a separate thread in our deployer service, for instance.

jali-clarke: out of scope of this pr, agreed

nri-kafka/src/Kafka.hs
```diff
@@ -281,12 +295,12 @@ sendHelperAsync producer doAnything onDeliveryCallback msg' = do
...
+            -- librdkafka invokes this callback exactly once per message, on
+            -- both success and failure, so handing it the whole delivery
+            -- report lets callers be notified of either outcome.
             ( \deliveryReport -> do
                 log <- Platform.silentHandler
-                Task.perform log <|
-                  case deliveryReport of
-                    Producer.DeliverySuccess _producerRecord _offset -> onDeliveryCallback
-                    _ -> Task.succeed ()
+                Task.perform log (onDeliveryReport deliveryReport)
```
jali-clarke, line 303: 💜

jali-clarke (approved):
thank you!

